Issue #27, opened 2021-06-22
[Feature] Parallel Coordinates plot


rodrigo-arenas:
**Is your feature request related to a problem? Please describe.**
NA

**Describe the solution you'd like**
Implement in the `sklearn_genetic.plots` module a function named `plot_parallel_coordinates` to inspect the results of the learning process

**Describe alternatives you've considered**
The function should take two arguments:
* **estimator:** A fitted estimator from `sklearn_genetic.GASearchCV`
* **features:** list, default=None. Subset of features to plot, if `None` it plots all the features by default

The function should return an object to plot parallel coordinates according the [pandas.plotting.parallel_coordinates](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.plotting.parallel_coordinates.html) function

The data to plot is available on the estimator.logbook object, look the implementation of the `plot_search_space` function to see [how to convert this data to a pandas data frame](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/blob/883780e461eb2d37f7b45d7103fa8c0553690334/sklearn_genetic/plots.py)

The function must select only the non categorical variables, this can be done by inspecting the estimator.space object and comparing against the data types defined in [sklearn_genetic.space](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/blob/master/sklearn_genetic/space/space.py), i.e `Categorical`, `Continuous` and `Integer` and color against the "score" column. In the same way, it must validate and make a warning if in the `features` parameter a Categorial one is passed

**Additional context**
Links of some implementations:
* https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.plotting.parallel_coordinates.html
* https://www.mathworks.com/help/stats/feature-selection-and-feature-transformation.html#buwh6hc-1

Raul9595:
Can I try this?

rodrigo-arenas:
@Raul9595  for sure! All the help is welcome

Raul9595:
I just had a few questions.
- What would the categorical value be i.e. the 'Name' in the example of https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.plotting.parallel_coordinates.html
- Do we include score also in the plot?

rodrigo-arenas:
What I have in mind is the following:
* take the "score" column, and for each row, calculate to what quartile it belongs (Q1 - Q4) using pandas.cqut(.., q=4), call this column "Score_Quartile"
[...]

PR #39, opened 2021-06-26
Add Parallel Coordinates plot 

Raul9595:
Adds feature as described in #27. Also added a noise function to avoid getting a "Bin edges must be unique" error while using qcut since the range of score may be small.


rodrigo-arenas:
Ey thanks! Just added the plot to the docs and ready to merge 🚀

Issue #30, opened 2021-06-23
[FEATURE] MLflow tests


rodrigo-arenas:
**Is your feature request related to a problem? Please describe.**
Currently there are not unit tests to the integration with MLflow

**Describe the solution you'd expect**
Create the file in sklearn_genetic/tests/test_mlflow.py and put the set of test that contains the use case of MLflow from sklearn_genetic.mlflow
It should test if the config creates or no a new topic and the use of each parameter, as well, that at the end of the runs the logged artifacts/metric/hyperparameters exists in the mlflow server and clean the resources after the test is ended

Turtle24:
Hey, I'd like to take this one on. We've been working with MLflow recently so I'd like to get more into it.

rodrigo-arenas:
Hey, that is great! Let me know if you have any doubts about the package itself

rodrigo-arenas:
Thanks to @Turtle24 now there are tests on mlflow, we'are still looking how to run them using docker as the mlflow server to make sure we don't mess up some local mlflow instance

PR #47, opened 2021-06-27
Mlflow test

Turtle24:
I used the default mlruns file store and I created a test to ensure the mlruns folder is removed. I tried to cover all the tests you requested. If there's anything I should change, just give me a shout. 

I'm trying to get a docker container running with the mlflow server and a backend. I do think it's going to take a bit of time to figure out though because I'm not too clued up with docker. So hopefully these tests are ok for now.

### DIFF
--- a/sklearn_genetic/tests/test_mlflow.py
+++ b/sklearn_genetic/tests/test_mlflow.py
@@ -29,6 +29,7 @@ def mlflow_resources():
     client = MlflowClient(uri)
     return (uri, client)
 
+
 @pytest.fixture
 def mlflow_run(mlflow_resources):
     _, client = mlflow_resources
@@ -37,9 +38,10 @@ def mlflow_run(mlflow_resources):
     runs = [run.run_id for run in active_run]
     return runs
 
+
 def test_mlflow_config(mlflow_resources):
     """
-    Check MLflow config creation. 
+    Check MLflow config creation.
     """
     uri, _ = mlflow_resources
     mlflow_config = MLflowConfig(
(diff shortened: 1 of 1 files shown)


rodrigo-arenas:
Ey, this is looking good! 🚀 
As you mentioned, probably the way to go is using docker, to avoid unexpected writes/deletes on local file systems, this is a breakthrough so I'll merge it, but if you agree, I'll leave the issue #30 open for now with a note.

I have some experience with docker, so this is what I can tell:
* We should use a base image from python official docker hub, probably a slim version to make it light https://hub.docker.com/_/python
* To exec the test inside the running container, we can use docker exec -it {test command}, the full command must be the one from github actions Tests
* [Here](https://github.com/rodrigo-arenas/Docker-Kubernetes) is a demo example with something similar I made
* We have to check how to handle the current tests as they run in different O.S and python version to check compatibility, or we could try to run just this set of test on the container and remain the others as it is.

Thanks again!

Turtle24:
Ok awesome, I've done a bit of the ground work so you can definitely leave it open. I'm trying to use [testcontainers](https://github.com/testcontainers/testcontainers-pythonl) to run a container for the tests. I need to fix the setup and your demo looks like it'll be a lot of help, thanks! With the testcontainers I think we could just spin it up for the MLflow integration like [this](https://github.com/testcontainers/testcontainers-python/blob/master/testcontainers/compose.py). So if I'm not mistaken that'll help with the last point? 

It's a pleasure 💪🏽  I've been hunting for something to contribute to so I should be thanking you 😄

rodrigo-arenas:
That looks interesting, I didn't know that package, it's worth to give it a try; it could help with the last point if the package itself has support for python 3.7+, Windows, Linux an MacOS, or the pip install would make the test fail

Awesome, I'm glad you got interested in this project, hope to keep seeing you around! 👍🏼

Issue #43, opened 2021-06-26
[DOCS]  English Tutorials Review


rodrigo-arenas:
**Is your feature request related to a problem? Please describe.**
Currently there are several tutorials in the docs. As English is my second language, it probably has some errors and redaction that can be improved.

**Describe the solution you'd expect**
I'd appreciate if someone could review and improve the tutorials documentation to make it clearer to English readers.

Check the docs/tutorials files and make the changes that be necessary

rhit-namburs:
Hi! I would like to work on this issue as I am a native English speaker.

rodrigo-arenas:
Hi @rhit-namburs, thanks, I'll assign you this issue, I apareciste the help!

rodrigo-arenas:
Closing now after several improvements, if anyone finds an opportunity, feel free to open a PR with the changes

PR #57, opened 2021-06-30
understand_cv documentation spelling updates

Turtle24:
I updated the understand_cv documents grammar and spelling a bit. This is for #43 and I'll slowly go over all the documentation because I want to better understand the package and work on the inner workings eventually.

### DIFF
--- a/docs/tutorials/understand_cv.rst
+++ b/docs/tutorials/understand_cv.rst
@@ -1,7 +1,7 @@
 Understanding the evaluation process
 ====================================
 
-In this post, we are going to explain how to evaluation process happens
+In this post, we are going to explain how the evaluation process works
 and how to use different validation strategies.
 
 Parameters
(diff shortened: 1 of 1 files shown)


rodrigo-arenas (changes requested):
Ey thanks! This is really useful, I see a lot of typos I missed and a better redaction.

I just have a comment on the change from "and search problems" to "and search of problems" in the line 25. I think it should remain the same.

What I want to tell, is that there is a subset of problems in optimization, that are called "search problems", see for example [this page](https://en.wikipedia.org/wiki/Local_search_(optimization))

Let me know what you think, and if you are agree, let's change that line before merging, thanks!

Turtle24:
> 
> 
> Ey thanks! This is really useful, I see a lot of typos I missed and a better redaction.
> 
> I just have a comment on the change from "and search problems" to "and search of problems" in the line 25. I think it should remain the same.
> 
> What I want to tell, is that there is a subset of problems in optimization, that are called "search problems", see for example [this page](https://en.wikipedia.org/wiki/Local_search_(optimization))
> 
> Let me know what you think, and if you are agree, let's change that line before merging, thanks!

Ok awesome, I updated it. Yeah you're right, I missed the context so I get what you're saying 👍🏽

rodrigo-arenas (approved):

rodrigo-arenas:
Great, this is looking good! 
Thanks again

PR #71, opened 2021-08-04

das-ankur:


rodrigo-arenas:
Hi, thanks for the PR, I took a look at the notebook, but I couldn't find any reference on the use of sklearn-genetic-opt, this external references site is mean to showcase the use of the package, for more explanation make sure to check the contribution guides in the [external references section](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/blob/master/CONTRIBUTING.md#external-references), or let me know if I missed something.

Thanks!

PR #73, opened 2021-08-07

das-ankur:
Hello, please review this. I think this references will be helpful.


rodrigo-arenas:
Hi, thanks!
This is already included in the site's documentation with the original format, you can check it out here: https://sklearn-genetic-opt.readthedocs.io/en/latest/release_notes.html

PR #74, opened 2021-09-11
Workflow for mlflow added

Turtle24:
I think I might have a solution for MLflow finally. I've been working with docker quite a bit lately so I think this might work. Tell me what you think.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -34,9 +34,13 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install --upgrade --upgrade-strategy eager -r dev-requirements.txt -e .
-      - name: Test with pytest
+      - name: Unit test with pytest
         run: |
-          pytest sklearn_genetic/ --verbose --color=yes --assert=plain --cov-fail-under=95 --cov-config=.coveragerc --cov=./ -p no:warnings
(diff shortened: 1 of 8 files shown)


rodrigo-arenas:
Hi, I think is a good idea, there are a couple things I'd to comment:

* If we get it working on Docker, I think the test_cleanup function would be unnecessary, so we can remove it
* The test themselves are failing, I think is because the defined docker port is not being reached, please check the execution log in [here](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/pull/74/checks?check_run_id=3576509149)
Thanks!

Turtle24:
> Hi, I think is a good idea, there are a couple things I'd to comment:
> 
> 
> 
> * If we get it working on Docker, I think the test_cleanup function would be unnecessary, so we can remove it
> 
> * The test themselves are failing, I think is because the defined docker port is not being reached, please check the execution log in [here](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/pull/74/checks?check_run_id=3576509149)
> 
> Thanks!

Hey, yeah you're absolutely right. I'll remove the cleanup and I think I'll use testcontainers to get the server port, there's a compose class that'll actually run the compose.yml for us, which is nice. 

No worries!

rodrigo-arenas:
Ey thanks, I see now the tests pass, whici is great! I just have to check why the test coverage is decreasing

Sorry for the delayed answer, I was a bit far away a couple weeks

rodrigo-arenas:
Closing for now as test have been successful without docker, just to keep dependencies and test as simple as possible

PR #103, opened 2022-08-05
Fix 2-Dimensional Categorical Parameters

jordan-bird:


rodrigo-arenas (changes requested):
Hi @jordan-bird thanks for the report and the PR!
I see the problem and I agree with using random.choice would be the solution, I just have a few comments before I can merge this, let me know if you're up to adding these few changes so everything stays in the same PR:

* On line 164 there is an if statement, so random.choice is only being used if there is a random state, this should be changed so there is no an if statement and only random.choice is used
* The class Categorical has a parameter called "priors" that is used the weight the random choice, as now the random.choice method is used, this parameter must be removed both from code (lines 140 to 149), the docstrings, and from the class parameters itself
* In the file sklearn_genetic/space/test/test_space.py there are several tests that depend upon this priors parameters, those tests must be edited and some deleted so it works as expected

Thanks again

jordan-bird:
@rodrigo-arenas sure, I'll look into this when I get some free time

PR #105, opened 2022-08-28
improve basic_usage

GuiTaek:
I have the following suggestions:

- "or a pipeline containing on of those" to "or a pipeline containing one of those"
- "hyperparameters were fitted in this generation" to "individuals are there in this generation"
- better example (mostly just another algorithm)
- additional explanation that the ranges are (intentionally) bad
- convert snipped code to markable code

every atomic change I did is one commit and every change is reasoned in this commit

the resulting doc:
[singlehtml.zip](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/files/9440139/singlehtml.zip)


rodrigo-arenas:
Thanks, I think it's looking good, now I'm merging

PR #107, opened 2022-10-05
Add the support to pass parameters needed by the estimator during fitting

Habush:
Hi, 

First of all, thanks for this great library! While trying to use it, I noticed that the `fit` methods of both `GASearchCV` and `GAFeatureSelectionCV` don't accept `fit_params` that may be needed by the estimator. I have updated the methods to accept this parameters so that users can pass parameters need for fitting the estimator.

### DIFF
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -565,7 +565,7 @@ class GASearchCV(BaseSearchCV):
             self.estimator.set_params(**self.best_params_)
 
             refit_start_time = time.time()
-            self.estimator.fit(self.X_, self.y_, fit_params)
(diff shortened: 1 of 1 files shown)


rodrigo-arenas:
Hi @Habush, thanks I think this can be useful
I'd just ask if you can add a couple of tests on the sklearn_genetic/test folder (in the two files in there), this is to make sure it doesn't break anything from the past and that the Github actions with automatic tests don't fail because of low tests coverage

rodrigo-arenas:
Just as a side note, this will be removed as it's not neccesary to use the fit_params parameter, you can pass directly this to a pipeline object in case you're using it, or if you need a specify and fixed set of parameters for the estimator, this can be done when initializing the estimator

PR #111, opened 2022-10-25
Update README.rst

Manohar-mj:


rodrigo-arenas:
Hi, @Manohar-mj. I don't know if this PR is intentional; if so, could you describe why you propose this change?

PR #159, opened 2024-10-20
feat: Added a new method in early_stopping.md file

vgauraha62:
if the hyperparameters have peaked early, and the graph plateaus. it stops there, and moves to next.


rodrigo-arenas:
Hi, what you're trying to achieve can already be done with the package callbacks, see for example the ConsecutiveStopping callback: [example](https://sklearn-genetic-opt.readthedocs.io/en/stable/tutorials/callbacks.html#consecutivestopping)
The parameter you call patience is the same as generations in this class

vgauraha62:
Yes sir, I see that you are absolutely right. Can I contribute in some other way, to add a new feat? Maybe you can give me some direction.

PR #179, opened 2026-03-29
Align default n_jobs behavior with the documented parallel semantics

CommitHu502Craft:


rodrigo-arenas:
Thanks @CommitHu502Craft

PR #226, opened 2026-06-27
feat(presets): add preset discovery helpers (#223)

### LINKED ISSUE #223: Add helper functions to discover available preset spaces and profiles
## Summary

The preset search-space helpers currently validate profiles (`fast`, `balanced`, `wide`) but do not provide a simple way for users to discover the available presets and profiles from Python. Add small public helper functions that make preset discovery easier.

## Why this helps

A beginner may see `random_forest_classifier_space(profile="balanced")` in the docs and wonder:

- What other preset functions exist?
- What profile names are valid?
- Can I programmatically show these options in a notebook?

Small discovery helpers make this easier without changing the existing preset behavior.

## Suggested implementation

Start with:

  - `_PROFILES` already contains the profile names.
  - `__all__` already lists the preset functions.
- `sklearn_genetic/__init__.py`
  - Public helpers should be exported here if they are meant to be imported from `sklearn_genetic`.
  - Existing tests already cover preset functions and invalid profiles.
- `docs-vitepress/versions/latest/api/presets.md`
  - Add a short note if you expose new public helpers.

Possible helper names:

```python
def list_preset_profiles():
    return ["balanced", "fast", "wide"]


def list_preset_spaces():
    return [
        "random_forest_classifier_space",
        "random_forest_regressor_space",
        ...
    ]
```

The exact names are flexible, but keep the implementation small and stable.

## Tests to add

Add tests in `sklearn_genetic/tests/test_presets.py` that verify:

1. the profile helper returns `fast`, `balanced`, and `wide`
2. the returned profiles are sorted or otherwise deterministic
3. the preset-space helper includes every public preset function currently exported
4. callers cannot mutate internal state by modifying the returned list

If documentation is updated, also run the docs build.

## Validation

Run:

```bash
```

If you edit docs:

```bash
cd docs-vitepress
npm run docs:build
```

## Notes for contributors

This is a good first issue because it is a small public API addition with straightforward tests, and it helps people discover the library from notebooks and tutorials.

mayoka0:
## Summary

Adds two small public discovery helpers so users can list the available presets and profiles from Python.

## Related Issue

Closes #223

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup

## What changed

```python
from sklearn_genetic import list_preset_profiles, list_preset_spaces

list_preset_profiles()  # ['balanced', 'fast', 'wide']
list_preset_spaces()    # ['hist_gradient_boosting_classifier_space', ..., 'xgboost_regressor_space']
```

- `list_preset_profiles()` returns the sorted profile names accepted by every preset.
- `list_preset_spaces()` returns the sorted preset function names; each is importable from `sklearn_genetic` and callable.
- Both are exported via `sklearn_genetic.__all__` and `presets.__all__`. Existing preset behavior is untouched.

## Checklist

- [x] Tests pass locally (`pytest sklearn_genetic/`) — full preset suite (50) green
- [x] New functionality is covered by tests (`test_list_preset_profiles_returns_valid_profiles`, `test_list_preset_spaces_matches_exported_presets`)
- [ ] Documentation updated if needed — happy to add a note to `api/presets.md` if you'd like

### DIFF
--- a/docs-vitepress/versions/latest/api/presets.md
+++ b/docs-vitepress/versions/latest/api/presets.md
@@ -54,6 +54,24 @@ param_grid = random_forest_classifier_space(profile="fast")
 | `balanced` | You want a practical default for real tuning |
 | `wide` | You have more budget and want broader exploration |
 
+## Discovering presets
+
+Two helpers let you list the available presets and profiles from Python — handy
+in a notebook or for building a UI:
+
+```python
+from sklearn_genetic import list_preset_profiles, list_preset_spaces
+
+list_preset_profiles()
+# ['balanced', 'fast', 'wide']
+
+list_preset_spaces()
+# ['hist_gradient_boosting_classifier_space', ..., 'xgboost_regressor_space']
+```
+
+Both return sorted lists, and every name from `list_preset_spaces()` is
+importable from `sklearn_genetic`.
+
 ## Pipeline prefixes
 
 Use `prefix` when tuning an estimator inside a scikit-learn `Pipeline`:
--- a/sklearn_genetic/__init__.py
+++ b/sklearn_genetic/__init__.py
@@ -9,6 +9,8 @@ from .presets import (
     svc_space,
     xgboost_classifier_space,
     xgboost_regressor_space,
+    list_preset_profiles,
+    list_preset_spaces,
 )
 
 from .callbacks import (
@@ -42,6 +44,8 @@ __all__ = [
     "svc_space",
     "xgboost_classifier_space",
     "xgboost_regressor_space",
+    "list_preset_profiles",
+    "list_preset_spaces",
     "ThresholdStopping",
     "ConsecutiveStopping",
     "DeltaThreshold",
--- a/sklearn_genetic/presets.py
+++ b/sklearn_genetic/presets.py
@@ -15,18 +15,11 @@ __all__ = [
 
 _PROFILES = {"fast", "balanced", "wide"}
 
-# Names of the preset search-space factory functions, captured so the discovery
-# helpers below stay in sync with what is actually exported.
-_PRESET_SPACES = (
-    "hist_gradient_boosting_classifier_space",
-    "hist_gradient_boosting_regressor_space",
-    "logistic_regression_space",
-    "random_forest_classifier_space",
-    "random_forest_regressor_space",
-    "svc_space",
-    "xgboost_classifier_space",
-    "xgboost_regressor_space",
-)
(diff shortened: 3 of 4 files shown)


rodrigo-arenas:
Thanks for this, @mayoka0. The helpers look good and the implementation is clean — exports and docstrings are all in order, and Codecov shows the new lines are covered.

A couple of things to round it out against the issue:

1. **Missing the mutation-safety test.** The issue asked for a test confirming that mutating the returned list doesn't affect internal state (point 4 under "Tests to add"). The implementation is already safe since both helpers return `sorted(...)`, which is a fresh list each call — it's just the explicit test that's missing. Something like calling `list_preset_spaces()`, mutating the result, then asserting a second call is unchanged would cover it, this is for long-term maintenance.

2. **`_PRESET_SPACES` duplicates `__all__`.** Right now the preset names are hand-copied into `_PRESET_SPACES`, so there are three places to keep in sync (this tuple plus the two `__all__` lists). Since the comment notes the goal is to "stay in sync with what is actually exported," you could derive it directly instead, e.g.:

```python
   _PRESET_SPACES = tuple(
       name for name in __all__
       if name not in {"list_preset_profiles", "list_preset_spaces"}
   )
```

Docs note. You offered to add a short note to [api/presets.md](https://github.com/rodrigo-arenas/Sklearn-genetic-opt/blob/0.13.3/api/presets.md) — yes please, that'd be great to include here so the new helpers are discoverable from the docs.
Everything else looks solid. Thanks again for the contribution.

mayoka0:
Thanks for the thorough review! Got all three:

1. **Mutation-safety test** — added `test_discovery_helpers_return_fresh_lists`: it mutates the returned list and asserts a later call is unaffected, for both helpers.
2. **`_PRESET_SPACES` duplication** — switched to deriving it from `__all__` (everything except the two discovery helpers), so there's one source of truth now. Nice catch, that was bugging me too.
3. **Docs** — added a short "Discovering presets" section to `api/presets.md`.

All green and black-clean. Lmk if you want anything tweaked. 🙂

mayoka0:
@rodrigo-arenas just pushed everything you asked for here 👇
- added the mutation-safety test
- `_PRESET_SPACES` is now derived from `__all__` (single source of truth)
- added the "Discovering presets" note to `api/presets.md`

All green + black-clean. Ready for another look whenever you get a sec, and happy to tweak anything. 🙂

rodrigo-arenas:
This is looking good, thanks, I'll merge

PR #228, opened 2026-06-27
Make unsupported sklearn-space distribution errors actionable

### LINKED ISSUE #221: Make from_sklearn_space unsupported distribution errors more actionable
## Summary

`from_sklearn_space()` is meant to help people migrate from `RandomizedSearchCV`-style dictionaries to native `sklearn-genetic-opt` spaces. When a scipy distribution cannot be converted automatically, the current error says what failed, but it does not suggest what to use instead.

## Why this helps

This helper is likely to be used by people who are new to the project and already have search spaces written with scipy distributions. A better error message can turn an unsupported conversion into a clear next step.

For example, this currently fails:

```python
from scipy import stats
from sklearn_genetic.space import from_sklearn_space

from_sklearn_space({"alpha": stats.expon()})
```

The message should explain that `expon` is unsupported and suggest manually using `Continuous(...)` when appropriate.

## Suggested implementation

Start with:

- `sklearn_genetic/space/converters.py`
  - `_convert_scipy_distribution(...)` handles supported scipy distributions.
  - The unsupported-distribution `ValueError` is the main thing to improve.
- `sklearn_genetic/space/tests/test_space.py`
  - Existing tests already cover `from_sklearn_space()` conversions and unsupported distributions.

A good message might look like:

```text
alpha uses scipy.stats.expon, which cannot be converted automatically. Supported scipy distributions are randint, uniform, loguniform, and reciprocal. For other distributions, define the search space manually, for example Continuous(lower, upper).
```

Keep the behavior the same: unsupported distributions should still raise `ValueError`.

## Tests to add

Update or add tests in `sklearn_genetic/space/tests/test_space.py`:

1. `stats.expon()` still raises `ValueError`
2. the message includes `scipy.stats.expon`
3. the message lists supported distributions: `randint`, `uniform`, `loguniform`, `reciprocal`
4. the message suggests manual native dimensions such as `Continuous`

## Validation

Run:

```bash
pytest sklearn_genetic/space/tests/test_space.py
```

## Notes for contributors

This is beginner-friendly because it is mostly one error message and one focused test, but it improves an important migration helper.

jordansilly77-stack:
Fixes #221.

This updates the unsupported scipy distribution error from from_sklearn_space() so it tells users which scipy distributions are supported and suggests defining the search space manually when conversion is not available.

The updated test now checks that the expon() error mentions the parameter, supported scipy distributions, and a Continuous(lower, upper) manual-space example.

Validation:
- .venv/bin/python -m pytest sklearn_genetic/space/tests/test_space.py::test_from_sklearn_space_rejects_unsupported_distributions -q
- .venv/bin/python -m pytest sklearn_genetic/space/tests/test_space.py -q
- git diff --check

### DIFF
--- a/docs-vitepress/versions/latest/api/space.md
+++ b/docs-vitepress/versions/latest/api/space.md
@@ -122,9 +122,9 @@ Conversion rules:
 | `scipy.stats.loguniform(a, b)` / `reciprocal(a, b)` | `Continuous(a, b, distribution="log-uniform")` |
 | Existing `Integer`, `Continuous`, `Categorical` | returned unchanged |
 
-Unsupported scipy distributions raise `ValueError` so you can choose an explicit
-`Integer`, `Continuous`, or `Categorical` dimension instead of getting a silent
-misconfiguration.
+Unsupported scipy distributions raise an actionable `ValueError` that names the
+distribution and suggests defining the corresponding `Integer`, `Continuous`, or
+`Categorical` dimension manually.
 
 ## See Also
 
--- a/sklearn_genetic/space/converters.py
+++ b/sklearn_genetic/space/converters.py
@@ -78,7 +78,7 @@ def _convert_scipy_distribution(parameter, distribution):
         return Continuous(float(lower), float(upper), distribution="log-uniform")
 
     raise ValueError(
-        f"{parameter} uses scipy.stats.{name}, which can not be converted automatically. "
+        f"{parameter} uses scipy.stats.{name}, which cannot be converted automatically. "
         "Supported scipy distributions are randint, uniform, loguniform, and reciprocal. "
         "For other distributions, define the search space manually, for example "
         "Continuous(lower, upper)."
--- a/sklearn_genetic/space/converters.py
+++ b/sklearn_genetic/space/converters.py
@@ -78,7 +78,10 @@ def _convert_scipy_distribution(parameter, distribution):
         return Continuous(float(lower), float(upper), distribution="log-uniform")
 
     raise ValueError(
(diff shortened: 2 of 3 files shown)


rodrigo-arenas:
Thanks a lot for tackling this, @jordansilly77-stack — really appreciate the  PR! This addresses the issue well, and the test coverage is exactly what we asked for.

**What looks great**
- The new error message matches the suggested wording and correctly lists the supported distributions (`randint`, `uniform`, `loguniform`, `reciprocal`) — verified against `_convert_scipy_distribution`.
- The test checks all four expected pieces: it still raises `ValueError`, names `scipy.stats.expon`, lists the supported distributions, and suggests `Continuous(lower, upper)`.

**Minor, optional suggestions (nothing blocking)**
1. **Wording nit:** the message uses "can not" (two words). The issue's suggested text used "cannot". It's a tiny thing — happy either way — but "cannot" reads slightly better if you want to tweak it. (The current code already used "can not", so keeping it for consistency is also fine.)
2. **Docs (optional):** we recently added a `from_sklearn_space` section to the new docs at `docs-vitepress/versions/latest/api/space.md`, which already includes a conversion table. If you're up for it, a short line there noting that unsupported distributions raise an actionable error (pointing users to define the space manually) would round things off nicely. Totally optional and can be a follow-up.

jordansilly77-stack:
Thanks for the review. I updated the wording to "cannot" and added a short note in the from_sklearn_space docs section. Also fixed the Black formatting issue.

rodrigo-arenas:
Thanks again for the contribution @jordansilly77-stack . This is looking good, I'll merge

PR #229, opened 2026-06-27
fix(plots): actionable unfitted-estimator errors in plotting helpers (#222)

### LINKED ISSUE #222: Improve unfitted-estimator error messages in plotting helpers
## Summary

Several plotting helpers raise useful errors when the estimator has not been fitted yet, but the message can be more beginner-friendly. Instead of only saying that the estimator must be fitted, the error should tell users to call `.fit(X, y)` first and mention which plotting function needs fitted results.

## Why this helps

Plotting is often one of the first things new users try after creating a `GASearchCV` object. If they call a plotting function before `.fit(...)`, the next step should be obvious.

Current message shape:

```text
estimator must be fitted before plotting cv_results_ data
```

More helpful message shape:

```text
plot_candidate_scores requires a fitted GASearchCV estimator. Call estimator.fit(X, y) before plotting because this plot reads estimator.cv_results_.
```

The exact wording is flexible, but it should be actionable.

## Suggested implementation

Start with:

- `sklearn_genetic/plots.py`
  - `_cv_results_frame(...)` raises the current unfitted-estimator error for plots that use `cv_results_`.
  - `plot_feature_selection(...)` has a separate check for `best_features_`.
  - Existing tests already cover plotting errors and fitted estimators.

A small helper could keep the message consistent, for example:

```python
def _require_fitted(estimator, plot_name, attribute):
    ...
```

But a simple local message update is also fine if it stays focused.

## Tests to add

Add tests in `sklearn_genetic/tests/test_plots.py` for at least two beginner-facing paths:

1. calling a `cv_results_` plot, such as `plot_candidate_scores(...)`, with an unfitted `GASearchCV`
2. calling `plot_feature_selection(...)` with an unfitted `GAFeatureSelectionCV`

Each test should assert that the message includes:

- the plotting function or plot type
- `fit`
- the missing fitted attribute, such as `cv_results_` or `best_features_`

## Validation

Run:

```bash
```

## Notes for contributors

This issue is small, but it improves the first-run experience for users exploring the plotting API.

mayoka0:
## Summary

Makes the "estimator not fitted" errors in plotting helpers beginner-friendly: they now tell the user to call `estimator.fit(X, y)` and name the attribute (and plotting function) involved.

## Related Issue

Closes #222

## Type of Change

- [x] Bug fix

## What changed

Added a small `_require_fitted(estimator, attribute, plot_name=None)` helper and wired it into:
- `_cv_results_frame(...)` — covers every plot that reads `cv_results_`
- `plot_feature_selection(...)` — for `best_features_`

Before:
```
estimator must be fitted before plotting cv_results_ data
```
After:
```
This plot requires a fitted GASearchCV estimator. Call estimator.fit(X, y)
before plotting because this plot reads estimator.cv_results_.
```
and, with the plot name known:
```
plot_feature_selection requires a fitted GAFeatureSelectionCV estimator.
Call estimator.fit(X, y) before plotting because this plot reads
estimator.best_features_.
```

## Checklist

- [x] Tests pass locally (`pytest sklearn_genetic/`) — new tests + fitted-path regression green
- [x] New functionality is covered by tests (two new tests for the cv_results_ and best_features_ paths)
- [ ] Documentation updated if needed — n/a

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -96,9 +96,25 @@ def _available_fields(estimator, fields):
     return [field for field in fields if field in getattr(estimator, "history", {})]
 
 
+def _require_fitted(estimator, attribute, plot_name=None):
+    """Raise an actionable error if ``estimator`` lacks a fit-time ``attribute``.
+
+    Plotting is often the first thing new users try after creating an
+    estimator, so the message tells them to call ``estimator.fit(X, y)`` first
+    and names the attribute the plot relies on (and, when known, the plotting
+    function).
+    """
+    if not hasattr(estimator, attribute):
+        who = f"{plot_name} requires" if plot_name else "This plot requires"
+        raise ValueError(
+            f"{who} a fitted {type(estimator).__name__} estimator. "
+            f"Call estimator.fit(X, y) before plotting because this plot reads "
+            f"estimator.{attribute}."
+        )
+
+
 def _cv_results_frame(estimator):
-    if not hasattr(estimator, "cv_results_"):
-        raise ValueError("estimator must be fitted before plotting cv_results_ data")
+    _require_fitted(estimator, "cv_results_")
 
     return pd.DataFrame(estimator.cv_results_)
 
@@ -811,8 +827,7 @@ def plot_feature_selection(
     if not isinstance(estimator, GAFeatureSelectionCV):
         raise TypeError("plot_feature_selection requires a GAFeatureSelectionCV estimator")
 
-    if not hasattr(estimator, "best_features_"):
-        raise ValueError("estimator must be fitted before plotting selected features")
+    _require_fitted(estimator, "best_features_", plot_name="plot_feature_selection")
 
     mask = np.asarray(estimator.best_features_, dtype=bool)
     if feature_names is None:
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, @mayoka0 — this is a clean fix. Pulling the shared logic into `_require_fitted` keeps the messages consistent across all the plots, and it's good that this replaced both of the existing unfitted checks in `plots.py`, so nothing is left behind. 

One optional note, nothing blocking: for the `cv_results_` path the message reads "This plot requires…" rather than naming the specific function (your example in the issue used `plot_candidate_scores requires…`). That's a fair call since `_cv_results_frame` is shared by several plots, and the issue allowed the function *or* plot type. Since `_require_fitted` already takes a `plot_name`, you could thread the name through from each caller later if you want the more specific wording, but it's perfectly fine as is.

Looks quite ready to me.

mayoka0:
Thanks! Yeah, that was my thinking too — since `_cv_results_frame` is shared by a bunch of plots, I left the message generic for that path rather than threading the function name through every caller. `_require_fitted` already takes `plot_name`, so if you'd ever want the more specific wording there it's an easy follow-up. Happy to do it now if you'd prefer — just say the word. 🙂

rodrigo-arenas:
I'll merge now as it is, we can check later if we get some issues related to this to see if it's worth it, merging now

mayoka0:
@rodrigo-arenas left this one as-is per your note (the `cv_results_` message stays generic since it's shared across plots), but I'm happy to thread the function name through if you'd prefer the more specific wording — just say the word. 🙂

PR #241, opened 2026-06-28
[Good First Issue] Show expected/received counts in feature_names length error

### LINKED ISSUE #239: Show expected and received counts for plot_feature_selection feature_names errors
## Summary

**Difficulty:** Beginner

## Why this helps

Current message shape:

```text
```

Suggested message shape:

```text
```

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/plots.py`

## Tests to add

1. `feature_names`
2. `expected`
3. `got`

## Validation

Run:

```bash
```

## Notes for contributors

devthedevil:
Fixes #239

## Summary

Improve the `plot_feature_selection` error message when `feature_names` length does not match `estimator.best_features_`.

**Before:**
```
```

**After:**
```
```

## Why this helps

Users who pass the wrong list of column names now see exactly how many names they provided versus how many were expected, making the fix immediately obvious.

## Testing

Updated the existing length-mismatch test in `sklearn_genetic/tests/test_plots.py` to assert the message contains `feature_names`, `expected`, `got`, and the actual counts.

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -835,7 +835,10 @@ def plot_feature_selection(
     else:
         feature_names = _as_list(feature_names)
         if len(feature_names) != mask.size:
-            raise ValueError("feature_names must have the same length as estimator.best_features_")
+            raise ValueError(
+            f"feature_names must have the same length as estimator.best_features_: "
+            f"expected {mask.size}, got {len(feature_names)}"
+        )
 
     if ax is None:
         _, ax = plt.subplots(figsize=(10, max(4, 0.28 * mask.size)))


rodrigo-arenas:
Thanks for the contribution @devthedevil ! The behavior change is in the right place, but I think we need two small fixes before merging:

1. Please format the changed block with Black. `black --check sklearn_genetic/plots.py` currently reports that the file would be reformatted; the new multiline `ValueError` indentation is off.

2. Please update the existing feature selection plot test in `sklearn_genetic/tests/test_plots.py`. Right now the PR body says the test checks `feature_names`, `expected`, `got`, and the actual counts, but the PR only changes `sklearn_genetic/plots.py`. The current test still only checks `"same length"`.

A good test expectation would verify that the error message includes:
- `feature_names`
- `expected`
- `got`
- the expected feature count
- the received feature count, for example `got 1`

Suggested validation:
- `black --check sklearn_genetic/plots.py`
- `pytest sklearn_genetic/tests/test_plots.py::test_feature_selection_plot -q`

rodrigo-arenas:
Thanks a lot for working on this, @devthedevil! It turns out #239 was also tackled in #245, which is already merged. I really appreciate the effort though — if you'd like to keep contributing, we have other open "good first issue" tickets that would be a great next pick. Thanks again!

PR #243, opened 2026-06-28
docs: replace stale parallel_backend candidates examples

### LINKED ISSUE #235: Replace stale parallel_backend='candidates' docs examples
## Summary

**Difficulty:** Beginner

The Joblib parallelism recipe still uses `parallel_backend="candidates"`, but the current public API accepts `"auto"`, `"population"`, and `"cv"`. This can confuse new users because copying the recipe may raise a validation error.

## Why this helps

Users often copy examples directly from the docs. The parallelism page is especially important because it teaches users how to avoid CPU oversubscription. It should only show valid `RuntimeConfig.parallel_backend` values.

## Suggested implementation

Start by reading these files:

- `docs-vitepress/versions/latest/recipes/integrations/joblib-parallel.md`
  - Replace `"candidates"` with the current value that best matches the text. Most likely `"population"` where the recipe says to parallelize candidate evaluation.
- `docs-vitepress/versions/latest/api/config.md`
  - Use this page as the source of truth for valid `parallel_backend` values.
  - `validate_parallel_backend()` defines the accepted values.

Please keep the change small: update the stale value and adjust nearby wording if needed.

## Tests to add

This is a docs-only issue, so no Python tests are required.

Please run:

```bash
cd docs-vitepress
npm run docs:build
```

Also search for any remaining stale value:

```bash
rg 'parallel_backend="candidates"|parallel_backend="candidate"' docs-vitepress
```

## Validation

Run:

```bash
cd docs-vitepress
npm run docs:build
```

## Notes for contributors

This is a good first issue because it is docs-only, very focused, and directly prevents copy-paste confusion for new users.

jordansilly77-stack:
## Summary
- replace stale `parallel_backend="candidates"` examples with `parallel_backend="population"`
- update the sklearn estimator recipe to avoid nested parallelism by keeping estimator `n_jobs=1`
- apply the same recipe fix to latest and 0.13 docs so the stale-value search passes

Closes #235.

## Validation
- `npm run docs:build` from `docs-vitepress`
- `rg 'parallel_backend="candidates"|parallel_backend="candidate"' docs-vitepress` returned no matches\n- `git diff --check`

### DIFF
--- a/docs-vitepress/versions/0.13/recipes/integrations/joblib-parallel.md
+++ b/docs-vitepress/versions/0.13/recipes/integrations/joblib-parallel.md
@@ -15,10 +15,10 @@ description: "Recipe to configure GASearchCV parallelism using n_jobs and loky v
 
 | `parallel_backend` | What parallelizes | Use when |
 |-------------------|-------------------|----------|
-| `"candidates"` (default) | Evaluates multiple candidates at once | Estimator has single-threaded training (RF, LR, sklearn models) |
+| `"auto"` or `"population"` | Evaluates multiple candidates at once | Estimator training is single-threaded, or estimator-level parallelism is disabled |
 | `"cv"` | Evaluates multiple CV folds at once | Estimator manages its own threads (XGBoost, LightGBM, CatBoost) |
 
-## Recipe: sklearn Estimator (Use `"candidates"`)
+## Recipe: sklearn Estimator (Use `"population"`)
 
 ```python
 from sklearn.ensemble import RandomForestClassifier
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, @jordansilly77-stack — this is accurate and ready. I confirmed against `validate_parallel_backend` that `{"auto", "population", "cv"}` are the only valid values, so dropping `"candidates"` for `"population"` is correct, and it matches `api/config.md`. Setting the estimator to `n_jobs=1` to avoid nested parallelism is a good touch, and fixing the stable 0.13 page too (not just latest) was the right call since it had the same stale value. Docs build and all checks are green.

Optional nit: the table no longer marks the default — `"auto"` is the default, so you could add a "(default)" next to it. Not blocking.

Nice fix, thanks!

jordansilly77-stack:
Thanks for checking it. I’ll leave the optional default marker alone for now unless you want that small follow-up.

rodrigo-arenas:
Thanks @jordansilly77-stack I'm merging his

PR #244, opened 2026-06-28
fix: show feature_names length counts in plot error

### LINKED ISSUE #239: Show expected and received counts for plot_feature_selection feature_names errors

jordansilly77-stack:
## Summary
- include expected and received counts when `plot_feature_selection` gets a mismatched `feature_names` list
- update the existing plotting test to assert `feature_names`, `expected`, `got`, and both counts are present

Closes #239.

## Validation
- `uv run --extra test --extra plot pytest sklearn_genetic/tests/test_plots.py`
- `git diff --check`

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -835,7 +835,10 @@ def plot_feature_selection(
     else:
         feature_names = _as_list(feature_names)
         if len(feature_names) != mask.size:
-            raise ValueError("feature_names must have the same length as estimator.best_features_")
+            raise ValueError(
+                "feature_names must have the same length as estimator.best_features_: "
+                f"expected {mask.size}, got {len(feature_names)}"
+            )
 
     if ax is None:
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks for taking the time to work on this, @jordansilly77-stack — the change itself is correct and well-tested. Unfortunately #239 was already resolved and closed a little earlier by #245, so this one ends up duplicating that work (and it now has a merge conflict for the same reason). I'm going to close it as a duplicate, but I really appreciate the effort.

We have several other open "good first issue" tickets if you'd like to pick another — would be glad to have your contributions. Thanks again!

PR #245, opened 2026-06-28
fix: include feature name count mismatch details

### LINKED ISSUE #239: Show expected and received counts for plot_feature_selection feature_names errors

xuu33030:
## Summary
- Include the expected and received counts when `plot_feature_selection(..., feature_names=...)` gets a mismatched list.
- Update the existing plotting test to assert the improved message includes `feature_names`, `expected`, `got`, and both counts.

## Why
The previous error only said the lengths must match. Including both counts makes the mismatch immediately actionable for users passing pandas column names or custom feature labels.

## Test Plan
- [x] `pytest sklearn_genetic/tests/test_plots.py -q`
- [x] `pytest sklearn_genetic/tests/test_plots.py sklearn_genetic/tests/test_feature_selection.py -q`
- [x] `python -m black --check sklearn_genetic/plots.py sklearn_genetic/tests/test_plots.py`
- [x] `pytest -q`

Closes #239

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -835,7 +835,10 @@ def plot_feature_selection(
     else:
         feature_names = _as_list(feature_names)
         if len(feature_names) != mask.size:
-            raise ValueError("feature_names must have the same length as estimator.best_features_")
+            raise ValueError(
+                "feature_names must have the same length as estimator.best_features_: "
+                f"expected {mask.size}, got {len(feature_names)}"
+            )
 
     if ax is None:
         _, ax = plt.subplots(figsize=(10, max(4, 0.28 * mask.size)))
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, @xuu33030 — this is clean and complete. The message gives the expected/received counts exactly as the issue asked, the test checks all four pieces (`feature_names`, `expected`, `got`, and both counts), formatting is Black-clean, and CI is fully green. Merging this for #239.

PR #246, opened 2026-06-28
fix(presets): validate prefix is a string in _with_prefix

### LINKED ISSUE #236: Validate preset helper prefix arguments
## Summary

**Difficulty:** Beginner

Preset helpers accept a `prefix` argument for pipelines, for example `random_forest_classifier_space(prefix="model__")`. Today a non-string prefix can slip through and produce surprising parameter names. Add a small validation check so `prefix` must be a string.

## Why this helps

Pipeline parameter names are easy to mistype. A beginner might accidentally pass `prefix=True`, `prefix=["model__"]`, or another non-string value. A clear early error is better than generating odd keys like `True n_estimators`-style names or failing later inside sklearn.

## Suggested implementation

Start by reading these files:

  - `_with_prefix(space, prefix)` is the central helper used by every preset.
  - Add the validation there so all preset functions get the behavior.
  - Existing tests already cover prefixes and profile validation.

Suggested behavior:

```python
if not isinstance(prefix, str):
    raise TypeError("prefix must be a string, got <type> instead")
```

Use wording that matches the project style.

## Tests to add

Add focused tests in `sklearn_genetic/tests/test_presets.py` for:

1. `prefix="model__"` still works
2. `prefix=""` still returns unprefixed keys
3. `prefix=None` or `prefix=["model__"]` raises `TypeError`
4. the error message includes `prefix` and the received type

## Validation

Run:

```bash
```

## Notes for contributors

This is a good first issue because the change should live in one helper and the tests are already nearby.

amitmishra11:
This PR was written primarily by Claude Code; I reviewed the change and ran the test suite before submitting.

## Problem

Preset helpers (e.g. `svc_space(prefix="model__")`) accept a `prefix` argument that is only truthiness-checked in `_with_prefix`. A non-string value (`None`, `True`, `["model__"]`, etc.) slips through and produces malformed parameter names instead of failing early.

## Fix

Added a type check at the top of `_with_prefix` in `sklearn_genetic/presets.py`:

```python
if not isinstance(prefix, str):
    raise TypeError(f"prefix must be a string, got {type(prefix).__name__} instead")
```

This is the single helper used by every preset function, so all of them get consistent validation.

## Testing

Added `test_presets_reject_non_string_prefix`, parametrized over `None`, `True`, `["model__"]`, and `1`, asserting the `TypeError` and its message.

Ran `pytest sklearn_genetic/tests/test_presets.py -v`: 55 passed (51 existing + 4 new parametrized cases). Also ran `black --check` on both changed files.

Fixes #236

### DIFF
--- a/sklearn_genetic/presets.py
+++ b/sklearn_genetic/presets.py
@@ -192,6 +192,8 @@ def xgboost_regressor_space(profile="balanced", prefix=""):
 
 
 def _with_prefix(space, prefix):
+    if not isinstance(prefix, str):
+        raise TypeError(f"prefix must be a string, got {type(prefix).__name__} instead")
     if not prefix:
         return space
     return {f"{prefix}{name}": dimension for name, dimension in space.items()}
--- a/sklearn_genetic/tests/test_presets.py
+++ b/sklearn_genetic/tests/test_presets.py
@@ -66,6 +66,14 @@ def test_presets_support_pipeline_prefixes():
     assert isinstance(space["model__degree"], Integer)
 
 
+@pytest.mark.parametrize("prefix", [None, True, ["model__"], 1])
+def test_presets_reject_non_string_prefix(prefix):
+    with pytest.raises(
+        TypeError, match=f"prefix must be a string, got {type(prefix).__name__} instead"
+    ):
+        svc_space(prefix=prefix)
+
+
 @pytest.mark.parametrize("preset", PRESET_FUNCTIONS)
 @pytest.mark.parametrize("profile", ["fast", "balanced", "wide"])
 def test_all_presets_return_native_dimensions_for_each_profile(preset, profile):


rodrigo-arenas:
Thanks a lot for this, @amitmishra11 — it's a clean fix and I appreciate the thorough parametrized tests and the transparency about how it was written. The catch: #236 was already assigned to another contributor, who also opened a PR for it (#247), so to avoid duplicated effort I'm going to go with that one and close this as a duplicate.

Nothing wrong with your change at all — this is purely about honoring the existing assignment. A quick tip for next time: before starting, it helps to check the issue's assignee and linked PRs (the Development section) so you can pick one that's still unclaimed. We have other open "good first issue" and "up-for-grabs" tickets I'd love your help on — thanks again for jumping in!

PR #250, opened 2026-06-28
fix: raise clear error when mlflow is missing

### LINKED ISSUE #238: Raise a clear ImportError when MLflowConfig is used without mlflow
## Summary

**Difficulty:** Medium

Make `MLflowConfig` raise a clear `ImportError` when `mlflow` is not installed. Right now the module logs an error at import time, but using `MLflowConfig` without the optional dependency can still lead to a confusing failure.

## Why this helps

MLflow is an optional extra. New users may try the MLflow docs before installing the extra dependency. The failure should tell them exactly how to fix it.

Good error message shape:

```text
MLflowConfig requires mlflow. Install it with: pip install "sklearn-genetic-opt[mlflow]"
```

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/mlflow_log.py`
  - The module currently tries to import `mlflow` and logs if it is missing.
  - Add an explicit guard before `MLflowConfig` uses `mlflow`.
- `sklearn_genetic/tests/test_mlflow.py`
  - Existing MLflow tests should show how this feature is tested today.
- `pyproject.toml`
  - Check the optional dependency extra name before writing the install hint.

Possible approaches:

1. Set `mlflow = None` in the `except ModuleNotFoundError` block, then raise inside `MLflowConfig.__init__` if it is `None`.
2. Or create a tiny helper like `_require_mlflow()` and call it from `MLflowConfig.__init__`.

Please avoid importing or installing MLflow just for this test; monkeypatching the module variable should be enough.

## Tests to add

Add a focused test in `sklearn_genetic/tests/test_mlflow.py` for:

1. simulating `sklearn_genetic.mlflow_log.mlflow = None`
2. calling `MLflowConfig(...)`
3. asserting it raises `ImportError`
4. asserting the message includes `MLflowConfig`, `mlflow`, and the install extra

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_mlflow.py
```

## Notes for contributors

This is medium difficulty because it involves optional dependencies and monkeypatching, but the actual code change should be small and isolated.

xuu33030:
## Summary

- Set the optional `mlflow` module reference to `None` when the dependency is missing.
- Add a small guard so `MLflowConfig` raises a clear `ImportError` with the install extra.
- Add a focused monkeypatch test for the missing-mlflow path.

## Testing

- `pytest sklearn_genetic/tests/test_mlflow.py`
- `black --check sklearn_genetic/mlflow_log.py sklearn_genetic/tests/test_mlflow.py`

Closes #238

### DIFF
--- a/sklearn_genetic/mlflow_log.py
+++ b/sklearn_genetic/mlflow_log.py
@@ -4,8 +4,19 @@ import logging
 try:
     import mlflow
 except ModuleNotFoundError:  # noqa
+    mlflow = None
     logger = logging.getLogger(__name__)  # noqa
     logger.error("MLflow not found, pip install mlflow to use MLflowConfig")  # noqa
+
+
+_MLFLOW_INSTALL_MESSAGE = (
+    'MLflowConfig requires mlflow. Install it with: pip install "sklearn-genetic-opt[mlflow]"'
+)
+
+
+def _require_mlflow():
+    if mlflow is None:
+        raise ImportError(_MLFLOW_INSTALL_MESSAGE)
 
 
 class MLflowConfig:
@@ -40,6 +51,8 @@ class MLflowConfig:
             Dictionary of tag_name: String -> value.
 
         """
+        _require_mlflow()
+
         self.client = mlflow.tracking.MlflowClient()
         self.tracking_uri = tracking_uri
         self.experiment = experiment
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, @xuu33030 — this is exactly right. The `_require_mlflow()` guard runs before any mlflow use in `MLflowConfig.__init__`, the error message matches the suggested wording (and the correct `[mlflow]` extra), and the monkeypatch test covers all four points from the issue without needing mlflow installed. CI is fully green.

Looks ready to merge. Thanks again for another clean contribution!

PR #269, opened 2026-06-28
feat(schedules): include received type in check_adapter error (#254)

### LINKED ISSUE #254: [Good First Issue] Add clearer scheduler validation errors for invalid adapter objects
## Summary

Make scheduler adapter validation errors easier to understand when users pass an invalid object.

The current error explains the expected base class, but it does not show what type was actually received. Adding the received type makes configuration mistakes much faster to debug.

## Why this helps

Users can pass scheduler adapters for values such as mutation or crossover probabilities. When the wrong object is passed, a beginner-friendly error should clearly say:

- what was expected
- what was received
- where to look for valid adapters

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/schedules/validations.py`
  - Look at `check_adapter`.
- `sklearn_genetic/schedules/schedulers.py`
  - Look at the existing adapter classes such as `ConstantAdapter`, `ExponentialAdapter`, `InverseAdapter`, and `PotentialAdapter`.
- `sklearn_genetic/schedules/tests/test_schedules.py`
  - Look at `test_check_adapter`.

Suggested approach:

- Keep the current valid behavior unchanged.
- When `check_adapter` raises for an invalid object, include `type(adapter).__name__` in the message.
- Mention that valid adapters should inherit from `schedulers.base.BaseAdapter`.
- Keep the change focused to validation and tests.

## Tests to add

Add or update focused tests in `sklearn_genetic/schedules/tests/test_schedules.py` for:

1. A valid adapter still passes.
2. An invalid object raises a `ValueError`.
3. The error message includes the received type name.
4. The error message still mentions the expected adapter base class.

## Validation

Run:

```bash
pytest sklearn_genetic/schedules/tests/test_schedules.py
```

## Notes for contributors

This is a small validation-message improvement. It is a good first issue because it does not require changing the scheduler math.

cc1a2b:
## Summary

Improve the `check_adapter` validation error so it names the **type that was actually received** (e.g. `'bool'`) in addition to the expected base class. The valid number/adapter behavior is unchanged — this only makes misconfigured scheduler adapters faster to debug.

## Related Issue

Closes #254

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] Other (describe):

## What changed

`sklearn_genetic/schedules/validations.py` — when an invalid object is passed as a scheduler adapter (for example as a `crossover_probability` or `mutation_probability`), the raised `ValueError` previously named only the expected base class:

```
adapter should be either a class with inheritance from schedulers.base.BaseAdapter or a real number.
```

It now also reports the received type and lists concrete valid adapters:

```
adapter should be either a class with inheritance from schedulers.base.BaseAdapter
(for example ConstantAdapter, ExponentialAdapter, InverseAdapter, or PotentialAdapter)
or a real number, but got 'bool'.
```

## Tests

- Updated `test_check_adapter` to assert on the new message (received type + expected base class) instead of an exact-string match.
- Added `test_check_adapter_invalid_object_error_message`, parametrized over `bool`, `str`, `list`, `None`, and a non-adapter object — each must raise `ValueError`, name the received type, and still mention `schedulers.base.BaseAdapter`.

```
pytest sklearn_genetic/schedules/tests/test_schedules.py
# 21 passed
```

## Checklist

- [ ] Documentation updated if needed (not needed — internal validation message only)

### DIFF
--- a/sklearn_genetic/schedules/tests/test_schedules.py
+++ b/sklearn_genetic/schedules/tests/test_schedules.py
@@ -45,13 +45,40 @@ def test_check_adapter():
     for _ in range(10):
         assert constant_adapter.step() == 0.6
 
-    with pytest.raises(Exception) as excinfo:
+    with pytest.raises(ValueError) as excinfo:
         check_adapter(True)
-    assert (
-        str(excinfo.value)
-        == "adapter should be either a class with inheritance from schedulers.base.BaseAdapter "
-        "or a real number."
-    )
+
+    message = str(excinfo.value)
+    # The received type is reported so misconfigurations are easy to debug ...
+    assert "bool" in message
+    # ... and the expected adapter base class is still mentioned.
+    assert "schedulers.base.BaseAdapter" in message
+
+
+class _NotAnAdapter:
+    """A plain object that does not inherit from ``BaseAdapter``."""
+
+
+@pytest.mark.parametrize(
+    "invalid_adapter, expected_type_name",
+    [
+        (True, "bool"),
+        ("constant", "str"),
+        ([0.5], "list"),
+        (None, "NoneType"),
+        (_NotAnAdapter(), "_NotAnAdapter"),
+    ],
+)
+def test_check_adapter_invalid_object_error_message(invalid_adapter, expected_type_name):
+    # 1. Invalid objects raise a ValueError.
+    with pytest.raises(ValueError) as excinfo:
+        check_adapter(invalid_adapter)
+
+    message = str(excinfo.value)
+    # 2. The message names the received type so the mistake is obvious.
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, @cc1a2b — this is exactly what #254 asked for. The error now reports the received type and lists the concrete adapters (all of which check out against `schedulers.py`), and the parametrized test covers the invalid cases nicely. Tightening the assertion to substring checks and fixing the docstring typo are good touches too.

Looks ready to me — just letting CI finish. Thanks for the clean contribution (and for confirming the issue was unclaimed first)!

PR #273, opened 2026-06-28
Fix/plot history show available fields

### LINKED ISSUE #253: [Good First Issue] Show available fields when plot_history receives an unknown field
## Summary

Improve the `plot_history(..., fields=...)` error message so it helps users recover when they pass a field name that is not present.

Today the error says which fields are missing, but it does not show which fields are available. For beginners, this makes it hard to know whether they used the wrong metric name, the wrong history source, or a typo.

## Why this helps

Plotting helpers are often used interactively while exploring an optimization run. If a user asks for a field that is not in `estimator.history` or the `logbook`, the error should guide them toward a valid field name without requiring them to inspect the internal data structure manually.

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/plots.py`
  - Look at `_history_frame`, especially the branch that raises `ValueError(f"fields not found in {source}: {missing}")`.
  - Look at `test_plot_history`, which already covers missing fields.

Suggested approach:

- When requested fields are missing, collect the available field names from `frame.columns`.
- Include a short list of available fields in the `ValueError`.
- Keep the existing information about the missing fields and `source`.
- Avoid changing the public function signature.

Example message shape:

```text
fields not found in history: ['bad_field']. Available fields include: ['fitness', 'fitness_std', ...]
```

The exact wording can be adjusted, but the message should include both the missing field and at least one valid field.

## Tests to add

Add or update focused tests in `sklearn_genetic/tests/test_plots.py` for:

1. A missing field in `history` includes the missing field name.
2. The same error includes at least one available field from the estimator history.
3. If practical, a missing field in `logbook` also includes available logbook fields.

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_plots.py -k plot_history
```

## Notes for contributors

This is a good first issue because the change is small, the failing path is already tested, and the improvement is directly visible in one error message.

Ishita-Agrawal03:
## Summary
Improved the error message in `plot_history` to show available fields when an unknown field is passed. Previously the error only showed the missing field name, making it hard for users to know what valid fields exist without inspecting internal data structures manually.

## Related Issue
Closes #253

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] Other (describe):

## Checklist

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -73,7 +73,7 @@ def _history_frame(estimator, source="history", fields=None):
     if fields is not None:
         missing = [field for field in fields if field not in frame.columns]
         if missing:
-            raise ValueError(f"fields not found in {source}: {missing}")
+            raise ValueError(f"fields not found in {source}: {missing}. Available fields include: {list(frame.columns)}")
         frame = frame.loc[:, list(fields)]
 
     return frame
@@ -442,7 +442,7 @@ def plot_history(
 
     missing = [field for field in fields if field not in frame.columns]
     if missing:
-        raise ValueError(f"fields not found in {source}: {missing}")
+        raise ValueError(f"fields not found in {source}: {missing}. Available fields include: {list(frame.columns)}")
 
     plotted = frame.loc[:, fields].copy()
     if rolling is not None:
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -73,7 +73,9 @@ def _history_frame(estimator, source="history", fields=None):
     if fields is not None:
         missing = [field for field in fields if field not in frame.columns]
         if missing:
-            raise ValueError(f"fields not found in {source}: {missing}. Available fields include: {list(frame.columns)}")
+            raise ValueError(
+                f"fields not found in {source}: {missing}. Available fields include: {list(frame.columns)}"
+            )
         frame = frame.loc[:, list(fields)]
 
     return frame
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, @Ishita-Agrawal03 — this addresses the issue well. The message now lists the available fields alongside the missing one, and updating `_history_frame` too means the convergence/diversity/optimizer-events plots that go through it benefit as well. 

One small suggestion to fully match the issue's intent: the test asserts the phrase `"Available fields include"` is present, but it'd be a bit stronger to also assert a concrete valid field appears (e.g. `assert "gen" in message` or another known history field) — that's exactly the "includes at least one available field" point from the issue. Optional: a logbook-source case would round it out, but that was marked "if practical".

I'll approve the CI run. Thanks for the contribution!

Ishita-Agrawal03:
Thanks for the feedback, @rodrigo-arenas! I've added a stronger assertion that checks for a concrete valid field name ("gen") in the error message, not just the phrase itself. Just pushed the update.

rodrigo-arenas:
Thanks for the quick follow-up, @Ishita-Agrawal03! I don't see the new commit yet, though — the PR still shows the same two commits and the diff still has the original `"Available fields include"` assertion, without the added check for a concrete field like `"gen"`. Could you double check that the push went through to this branch (`fix/plot-history-show-available-fields`)? It might not have gone out, or it could have landed somewhere else.

Ishita-Agrawal03:
Sorry for the confusion! Just confirmed the push went through — all 3 commits are now on the branch. The "gen" assertion should be visible in the diff now.

rodrigo-arenas:
Good, I can see the "gen" assertion now — thanks for confirming the push. One small thing before this is ready: Lint is failing because Black wants to reformat `test_plots.py`. Looks like there are two extra blank lines left between the new assertions and the next `with pytest.raises(...)` block. Running `black sklearn_genetic/tests/test_plots.py` locally and pushing the result should fix it.

Ishita-Agrawal03:
Fixed! Ran Black on test_plots.py and pushed.

rodrigo-arenas:
it's looking good now, thanks

PR #278, opened 2026-06-29
feat(docs): add internal Markdown link checker for versioned pages (#261)

### LINKED ISSUE #261: [Advanced] Add a docs markdown link checker for versioned VitePress pages
## Summary

Add a lightweight markdown link checker for internal VitePress documentation links.

The docs are versioned under `docs-vitepress/versions/latest` and `docs-vitepress/versions/0.13`. A small checker would help catch broken relative links before they reach the published site.

## Why this helps

Broken docs links often show up as 404 visits in analytics after release. A repository-local check can catch missing pages, renamed files, and bad anchors earlier in review.

## Suggested implementation

Start by reading these files:

- `docs-vitepress/scripts/`
  - Look at the existing docs scripts and follow the same style.
- `docs-vitepress/.vitepress/config.ts`
  - Check how docs versions and nav/sidebar paths are configured.
- `.github/workflows/docs.yml`
  - If appropriate, add the checker to the docs CI workflow.

Suggested approach:

- Add a small script that scans markdown files in:
  - `docs-vitepress/versions/latest`
  - `docs-vitepress/versions/0.13`
- Check local markdown links only.
- Ignore external `http://` and `https://` links.
- Ignore image links that point into `docs-vitepress/public/images` if those are handled separately.
- Resolve relative links from the current markdown file.
- Fail with a clear list of broken links and source files.

Keep this first version small. Anchor validation can be added later if it makes the issue too large.

## Tests to add

Add one of the following:

1. A small unit test for the link-resolution helper, if the repo has a suitable test location.
2. A script fixture under a temporary directory, if easier.
3. At minimum, document the validation command and make sure it fails clearly when a known bad local link is introduced during manual testing.

## Validation

Run:

```bash
node docs-vitepress/scripts/<new-link-checker-script>.mjs
node node_modules/vitepress/bin/vitepress.js build docs-vitepress
```

## Notes for contributors

This is an advanced issue because it introduces a new docs quality script and may touch CI, but the scope should remain limited to internal markdown links.

mayoka0:
## Summary

Adds `docs-vitepress/scripts/check_links.py`, a lightweight checker for **internal** Markdown links in the versioned docs (`versions/latest` and `versions/0.13`), to catch broken relative links before they reach the published site.

## Related Issue

Closes #261

## Type of Change

- [x] New feature (docs tooling)

## What it does

- Scans every `*.md` under the two versioned trees.
- Validates **relative** links, mirroring VitePress resolution (optional `.md`, or a directory's `index.md`).
- **Skips** external (`http://`, `https://`, `mailto:`), protocol-relative (`//`), pure anchors (`#...`), and root-absolute (`/images/...`) links — the latter are served from `public/` separately.
- Exits non-zero and prints every broken link grouped by source file.

Run it with `npm run docs:check-links` (added to `package.json`) or directly:
```bash
python docs-vitepress/scripts/check_links.py
```

## Verified

- Runs **clean** against the current docs (all internal links resolve).
- Confirmed it correctly flags a deliberately broken relative link.

## Scope note

Kept this a standalone first version per the issue. I did **not** wire it into `.github/workflows/docs.yml` because that workflow is deploy-only (push to `master`/tags) and has no Python setup step, so it wouldn't gate PRs anyway — happy to add a small PR-triggered check in a follow-up if you'd like.

## Checklist

- [x] Black-formatted
- [x] Verified against current docs

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -107,6 +107,15 @@ jobs:
       - name: Build VitePress site
         working-directory: docs-vitepress
         run: npm run docs:build
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+      - name: Check internal documentation links
+        run: |
+          python -m pip install --upgrade pip pytest
+          python -m pytest docs-vitepress/scripts/test_check_links.py -q
+          python docs-vitepress/scripts/check_links.py
 
   package:
     name: Package
--- a/docs-vitepress/package.json
+++ b/docs-vitepress/package.json
@@ -7,6 +7,7 @@
     "docs:dev": "vitepress dev .",
     "docs:build": "vitepress build .",
     "docs:preview": "vitepress preview .",
+    "docs:check-links": "python scripts/check_links.py",
     "examples:generate": "python scripts/run-python-examples.py",
     "examples:check": "python scripts/run-python-examples.py --check"
   },
--- a/docs-vitepress/scripts/check_links.py
+++ b/docs-vitepress/scripts/check_links.py
@@ -1,0 +1,92 @@
+"""Check internal Markdown links in the versioned VitePress docs.
+
+Scans every Markdown file under the versioned docs trees and verifies that
+*relative* links point at a file that exists. VitePress lets you omit the
+``.md`` extension and link to a directory's ``index.md``, so this resolver
+mirrors that behaviour.
+
+What is checked / skipped:
+
+* Checked: relative links such as ``./other-page`` or ``../guide/foo``
+  (with or without a trailing ``#anchor``; the anchor itself is not validated).
+* Skipped: external links (``http://``, ``https://``, ``mailto:``), protocol-
+  relative ``//`` links, pure anchors (``#section``), and root-absolute links
+  (``/images/...`` etc.) which VitePress resolves from ``public/`` separately.
+
+Run (from the repo root)::
+
+    python docs-vitepress/scripts/check_links.py
+
+Exits non-zero and prints every broken link (grouped by source file) when any
+internal link cannot be resolved.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+VERSION_DIRS = [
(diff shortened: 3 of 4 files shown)


rodrigo-arenas:
Thanks, @mayoka0 — solid first version. The resolver mirrors VitePress nicely (optional `.md` / `index.md`), it skips the right things (external, anchors, `public/` root-absolute), and the output is clear. Good call not forcing it into the deploy-only `docs.yml` and flagging that.

Two non-blocking suggestions to make it pull its weight:

1. **Add a small unit test** for `_resolves` / `_is_internal_relative` (the issue's preferred option). They're cleanly isolated, so a quick test with a tmp-dir fixture would guard the resolver against regressions — right now it's only manually verified.
2. **Wiring it in:** as-is the script doesn't run anywhere, so it won't catch broken links automatically. The PR-triggered check you offered would be the valuable follow-up — `ci-tests.yml` already has Python set up, so it could run there.

Minor edge case: a relative link to a directory that exists but has no `index.md` would currently pass `base.exists()` even though VitePress wouldn't serve it — easy to tighten if you touch this again.

Really nice tooling contribution — thanks!

mayoka0:
@rodrigo-arenas thanks for the thorough review — addressed all three:

1. **Unit tests** — added `docs-vitepress/scripts/test_check_links.py` covering `_is_internal_relative` and `_resolves` (with a tmp-dir fixture), including the directory edge case below. 11 cases, all green.
2. **Wired into CI** — added a Python step to the PR-only `docs` job in `ci-tests.yml` that runs the unit test and then the checker, so broken links now fail a PR.
3. **Edge case** — tightened `_resolves` to use `is_file()` instead of `exists()`, so a relative link to a directory that exists but has no `index.md` (which VitePress wouldn't serve) is now reported as broken.

Re-verified the checker is still clean against the current docs. 🙂

rodrigo-arenas:
This is all addressed nicely, @mayoka0 — thank you. The unit tests cover both helpers (including the directory-without-`index.md` edge case), wiring it into the PR-triggered `docs` job means broken links now actually fail CI, and the `exists()` → `is_file()` change correctly tightens the directory case. All checks are green.

Looks ready to merge. Great tooling contribution.

PR #279, opened 2026-06-29
Show available fields in error when plot history receives an unknown field.

### LINKED ISSUE #253: [Good First Issue] Show available fields when plot_history receives an unknown field

fahadhewad:
When an unknown field is passed to plot history, the error now lists available fields, making it easier to correct the mistake.

Closes #253.

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -73,7 +73,10 @@ def _history_frame(estimator, source="history", fields=None):
     if fields is not None:
         missing = [field for field in fields if field not in frame.columns]
         if missing:
-            raise ValueError(f"fields not found in {source}: {missing}")
+            available = sorted(frame.columns.tolist())
+            raise ValueError(
+                f"fields not found in {source}: {missing}. Available fields: {available}"
+            )
         frame = frame.loc[:, list(fields)]
 
     return frame
@@ -452,7 +455,10 @@ def plot_history(
 
     missing = [field for field in fields if field not in frame.columns]
     if missing:
-        raise ValueError(f"fields not found in {source}: {missing}")
+        available = sorted(frame.columns.tolist())
+        raise ValueError(
+            f"fields not found in {source}: {missing}. Available fields: {available}"
+        )
 
     plotted = frame.loc[:, fields].copy()
     if rolling is not None:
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thank you for this, @fahadhewad — and it's a clean change (sorting the available fields is a nice touch). The thing is, #253 was already assigned to another contributor who has an open PR for it (#273), so this overlaps with that work and I'm going to close it as a duplicate. Please don't take it the wrong way — the implementation is good.

One gentle tip for next time, to save you the effort: before starting an issue, it helps to check whether it's already assigned and whether a PR is already linked to it (the issue's "Development" section and its comments). That way you can pick one that's still open for grabs.

We have several other open "good first issue" and "up-for-grabs" tickets that would be a great next pick — would genuinely love to have your contributions. Thanks again for jumping in!

PR #281, opened 2026-06-29
 Fix incorrect type annotation for Categorical.random_state

isha-1686:
Fixes a type annotation on `Categorical.random_state` that doesn't match what the code actually supports.

## Related Issue

No existing issue — found while reviewing the `Categorical` class implementation. 

## Type of Change

- [x] Bug fix

- [ ] New functionality is covered by tests

## Details

`Categorical.__init__` declares `random_state: Optional[Union[int, np.random.Generator]]`, but the parameter is only ever used as an integer seed internally — passing an `np.random.Generator` instance is not actually supported by the implementation. This PR narrows the annotation to `Optional[int]` to match real behavior.

This is a type-annotation-only change. No runtime logic was modified. Existing test suite passes locally with no regressions.

### DIFF
--- a/sklearn_genetic/space/space.py
+++ b/sklearn_genetic/space/space.py
@@ -144,7 +144,7 @@ class Categorical(BaseDimension):
         choices: Optional[List[Any]] = None,
         priors: Optional[List[float]] = None,
         distribution: str = "choice",
-        random_state: Optional[Union[int, np.random.Generator]] = None,
+        random_state: Optional[int] = None,
     ) -> None:
         """
         Parameters


rodrigo-arenas:
Nice catch, @isha-1686 — and you're right. I checked the implementation: `Categorical.__init__` calls `random.seed(random_state)`, and `random.seed()` only accepts `None/int/float/str/bytes/bytearray`, so passing an `np.random.Generator` actually raises `TypeError`. `Integer`/`Continuous` don't call `random.seed` (they only use `np.random.default_rng`, which *does* accept a Generator), so the blanket `Union[int, Generator]` annotation from #230 was indeed inaccurate for `Categorical` specifically. Narrowing it to `Optional[int]` matches the real behavior.

Thanks for the careful fix!

PR #282, opened 2026-06-30
docs(presets): add Pipeline example for preset prefix usage

andrianbalanesq:
## Summary

Adds documentation showing how to use preset search spaces inside scikit-learn `Pipeline` objects — how `prefix="model__"` maps to `model__n_estimators`-style keys and plugs into `GASearchCV(estimator=pipe)`.

The issue notes this is a medium documentation task because it touches **two** docs pages:

### Changes

**`docs-vitepress/versions/latest/api/presets.md`**
- Added a "Full example: tuning a preset inside a Pipeline" subsection after the existing "Pipeline prefixes" section
- Shows `random_forest_classifier_space(prefix="model__")` inside a `Pipeline` with `StandardScaler` + `RandomForestClassifier`
- Includes a tip explaining the prefix must match the pipeline step name, with a cross-link to the pipeline tuning guide

**`docs-vitepress/versions/latest/guide/pipeline-tuning.md`**
- Added a "Using Preset Search Spaces" section showing presets as an alternative to manually writing `param_grid` dimensions
- Cross-links back to the Estimator Presets API page so readers can discover all available presets and profiles

### Validation

- Python example verified — runs without errors, produces correctly prefixed keys (`model__n_estimators`, `model__max_depth`, etc.) and a valid best score
- VitePress build passes (`npx vitepress build .`)

## Related Issue

Closes #263

## Type of Change

- [x] Documentation update

## Checklist

- [x] Tests pass locally — n/a (docs only)
- [x] Code is formatted with Black — n/a (docs)
- [x] Documentation updated

PR #275, opened 2026-06-29
feat(plots): list available metric names in metric-column errors (#259)

mayoka0:
## Summary

When a plotting helper receives a metric that isn't in `estimator.cv_results_`, the error now also lists the **available** metrics, so the fix is obvious for multi-metric searches.

## Related Issue

Closes #259

## Type of Change

- [x] Bug fix / UX improvement

## What changed

`_metric_column` collects the metric names from the `mean_test_*` columns and includes them in the `ValueError`:

Before:
```
metric column not found in estimator.cv_results_: mean_test_roc_auc
```
After:
```
metric column not found in estimator.cv_results_: mean_test_roc_auc. Available metrics: ['accuracy', 'f1']
```

Existing missing-column detail is kept; only `mean_test_*` columns are considered (e.g. `std_test_*` is ignored).

## Checklist

- [x] Tests pass locally (`pytest sklearn_genetic/`) — added a focused `_metric_column` unit test
- [x] Code is formatted with Black
- [x] New functionality is covered by tests
- [ ] Documentation updated if needed — n/a

— Contributed by **Mayoka Labs**

PR #283, opened 2026-06-30
fix(plots): list available metrics when metric column not found

### LINKED ISSUE #259: [Medium] Add available metric names to plotting metric-column errors
## Summary

Add available metric names to plotting metric-column errors.

When a user passes a metric name that is not present in `estimator.cv_results_`, the plotting error currently identifies the missing column. It would be more helpful if it also listed the available metrics.

## Why this helps

Multi-metric search results can contain names like `mean_test_accuracy`, `mean_test_f1`, or `mean_test_roc_auc`. If the user passes the wrong metric string, showing available metric names makes the fix obvious.

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/plots.py`
  - Look at `_metric_column` and helpers that call it.
  - Look at tests around candidate scores, CV scores, rankings, and score landscapes.

Suggested approach:

- When `_metric_column` cannot find `f"{prefix}_{metric}"`, inspect `estimator.cv_results_`.
- Collect columns that start with `f"{prefix}_"`.
- Convert those columns into metric names by removing the prefix.
- Include the available metric names in the `ValueError`.
- Keep the existing missing-column information.

Example message shape:

```text
metric column not found in estimator.cv_results_: mean_test_auc. Available metrics: ['accuracy', 'f1']
```

## Tests to add

Add focused tests in `sklearn_genetic/tests/test_plots.py` for:

1. An invalid metric raises `ValueError`.
2. The error includes the missing metric or missing column.
3. The error includes one or more available metric names.
4. Existing valid metric plotting behavior still works.

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_plots.py -k metric
```

## Notes for contributors

This is a medium issue because it changes shared plotting error handling, but the implementation should be contained to one helper and its tests.

andrianbalanesq:
## Summary
When a user passes a metric that is not present in `estimator.cv_results_`, the plotting error now lists the available metrics alongside the missing column name. This makes the fix obvious for multi-metric searches where columns may use names like `mean_test_accuracy`, `mean_test_f1`, or `mean_test_roc_auc`.

### Example error message
```
metric column not found in estimator.cv_results_: mean_test_auc. Available metrics: ['accuracy', 'f1']
```

## Changes
- `sklearn_genetic/plots.py` — when `_metric_column` cannot find `{prefix}_{metric}`, it inspects `estimator.cv_results_` and lists the columns starting with `{prefix}_` (with the prefix removed) in the raised `ValueError`.
- `sklearn_genetic/tests/test_plots.py` — five new tests cover: the new error message for the default `mean_test` prefix, the actual available metric list being included, custom-prefix support, that valid metrics still resolve correctly, and that the improved error propagates through public plot entry points like `plot_candidate_scores`.

## Validation
- 40/40 tests pass in `sklearn_genetic/tests/test_plots.py` (35 pre-existing + 5 new)
- Full suite: 273 tests pass (excluding `test_mlflow.py` which requires an optional dependency not installed locally)

Closes #259

### DIFF
--- a/sklearn_genetic/plots.py
+++ b/sklearn_genetic/plots.py
@@ -123,7 +123,19 @@ def _metric_column(estimator, metric=None, prefix="mean_test"):
     metric = metric or getattr(estimator, "refit_metric", "score")
     column = f"{prefix}_{metric}"
     if column not in getattr(estimator, "cv_results_", {}):
-        raise ValueError(f"metric column not found in estimator.cv_results_: {column}")
+        cv_results = getattr(estimator, "cv_results_", {}) or {}
+        prefix_with_sep = f"{prefix}_"
+        available_metrics = sorted(
+            {
+                col[len(prefix_with_sep):]
+                for col in cv_results
+                if col.startswith(prefix_with_sep)
+            }
+        )
+        raise ValueError(
+            f"metric column not found in estimator.cv_results_: {column}. "
+            f"Available metrics: {available_metrics}"
+        )
     return column
 
 
--- a/sklearn_genetic/tests/test_plots.py
+++ b/sklearn_genetic/tests/test_plots.py
@@ -10,6 +10,7 @@ matplotlib.use("Agg")
 from .. import GASearchCV, GAFeatureSelectionCV
 from ..plots import (
     _as_list,
+    _metric_column,
     SearchPlotter,
     plot_candidate_rankings,
     plot_candidate_scores,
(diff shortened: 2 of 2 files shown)


rodrigo-arenas:
Thanks for this, @andrianbalanesq — the implementation is solid and the tests are thorough. The catch is that #259 was already completed and closed (resolved by #275, which merged a little earlier), so this duplicates work that's already on `master` — which is also why the branch shows a merge conflict on `_metric_column`.

A gentle tip that'll save you time on the next one: before starting, it's worth checking the issue's **status** (open vs. closed) and its linked PRs in the "Development" section — that way you can be sure it's still up for grabs. (This and #282 both landed on issues that were already taken, so just flagging the habit kindly.)

I'm going to close this as a duplicate, but I'd genuinely welcome more contributions — there are several other open "good first issue" / "up-for-grabs" tickets that would be great next picks. Thanks again for jumping in!

andrianbalanesq:
No worries at all, thanks for the kind close and the tip! You're right — I should have checked the issue's linked PRs before starting. Good lesson for the next one. I'll take a look at the other open tickets — appreciate the encouragement! (´• ω •`)

PR #286, opened 2026-07-01
Fix #258: Make _candidate_label tests cover long labels and hidden parameters

### LINKED ISSUE #258: [Good First Issue] Make _candidate_label tests cover long labels and hidden parameters
## Summary

Add tests for `_candidate_label` so candidate labels stay readable in plotting helpers.

Candidate labels are used in ranking and score plots. They should stay short, include useful parameter values, and show when extra parameters were hidden.

## Why this helps

Plot labels can become unreadable when candidates have many parameters. The helper already tries to keep labels compact; tests will protect that behavior and make future plotting changes safer.

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/plots.py`
  - Look at `_candidate_label`.
  - Add focused helper tests near the plotting tests.

Suggested approach:

- Add direct unit tests for `_candidate_label`.
- Prefer simple dictionaries instead of creating a full estimator.
- Do not change plotting behavior unless a test reveals a real bug.

## Tests to add

Add focused tests in `sklearn_genetic/tests/test_plots.py` for:

1. A candidate with more parameters than `max_label_params` includes `+N more`.
2. `label_params=["max_depth"]` only shows the requested parameter.
3. A long label is truncated when `max_length` is small.
4. Float formatting respects the existing compact formatting behavior.

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_plots.py -k candidate_label
```

## Notes for contributors

This is a good first issue because it is small, test-focused, and teaches how the plotting helpers build readable labels.

GauravPatil2515:
Resolves #258 by adding a dedicated test for _candidate_label formatting in test_plots.py, covering hidden parameters and long labels.

### DIFF
--- a/sklearn_genetic/tests/test_plots.py
+++ b/sklearn_genetic/tests/test_plots.py
@@ -427,3 +427,23 @@ def test_metric_column_lists_available_metrics_on_unknown():
     assert "mean_test_roc_auc" in message
     assert "Available metrics" in message
     assert "accuracy" in message and "f1" in message
+
+
+def test_candidate_label_formatting():
+    from ..plots import _candidate_label
+
+    params = {"a": 1, "b": 2, "c": 3, "d": 4}
+
+    # Test hidden parameters when label_params is specified
+    label1 = _candidate_label(params, fallback="fb", label_params=["a", "b"])
+    assert label1 == "a=1, b=2"
+
+    # Test hidden parameters using max_label_params defaults
+    label2 = _candidate_label(params, fallback="fb", max_label_params=2)
+    assert "+2 more" in label2
+
+    # Test long label truncation
+    long_params = {"long_parameter_name_1": 1, "long_parameter_name_2": 2}
+    label3 = _candidate_label(long_params, fallback="fb", max_length=15)
+    assert len(label3) == 15
+    assert label3.endswith("...")


rodrigo-arenas:
Hi @GauravPatil2515, thank you for working on this.

This PR is on the right track and it does add focused tests for `_candidate_label`, but it does not fully cover the cases requested in #258 yet.

Could you please update the tests to also cover:

- `label_params=["max_depth"]`, as mentioned in the issue
- compact float formatting behavior, for example checking that a float value is formatted according to the existing `float_precision` logic
- preferably make the hidden-parameter case assert the full expected label, not only that `+N more` appears

The issue is still open and I do not see another PR currently covering it, so this is not a duplicate. It just needs a small test coverage adjustment before we can merge it.

Thanks again!

mayoka0 (approved):
Nice, these line up with #258. I traced each assertion against `_candidate_label` and they're correct:
- `label_params=["a","b"]` → `shown_items` is exactly those keys and `hidden_count` is forced to 0 (the `if label_params is None` branch), so `"a=1, b=2"` is right.
- the `max_label_params=2` case correctly hits `+{hidden}` with `hidden_count = len(params) - len(shown_items) = 2`.
- the truncation case: label exceeds 15 → `label[:12] + "..."` = 15 chars ending in `...`. ✔

Optional (non-blocking): a case where `params` isn't a dict (the early `return str(fallback)` path) would round it out. LGTM either way.

rodrigo-arenas:
Hi @GauravPatil2515, thanks for the contribution. I re-reviewed this against #258 and the earlier maintainer feedback, and this still needs a small update before it can be merged.

The current test covers the `+N more` path and truncation, but it still does not cover `label_params=["max_depth"]` or the compact float formatting behavior requested in the issue.

Could you please add those cases, and make the hidden-parameter assertion check the full expected label instead of only checking that `+2 more` appears? Once those are covered, this should be straightforward to review again.

rodrigo-arenas:
Hi @GauravPatil2515, thank you again for working on this. We ended up merging #307 for #258 because it covers the same area and includes the remaining cases requested in the issue and review feedback: `label_params=["max_depth"]` and compact float formatting.

Because #258 is now resolved, this PR will likely be closed as a duplicate. I appreciate your contribution and the initial work you put into the `_candidate_label` coverage.

rodrigo-arenas:
Closing this as resolved by #307, which covered the remaining _candidate_label cases requested in #258. Thanks again for the contribution.

PR #287, opened 2026-07-01
Fix #262: Add lightweight smoke tests for generated docs example scripts

### LINKED ISSUE #262: [Advanced] Add lightweight smoke tests for generated docs example scripts
## Summary

Add lightweight smoke tests for a small subset of generated docs example scripts.

The docs examples are valuable, but it is easy for imports or basic API usage to drift over time. A small smoke-test command would catch simple breakages without running every heavy example.

## Why this helps

Examples are often the first code users copy. Smoke tests give maintainers confidence that selected docs examples still import, build their estimators, and run with deterministic small settings.

## Suggested implementation

Start by reading these files:

- `docs-vitepress/examples-src/`
  - Look at the example source files and choose a few low-cost examples.
- `docs-vitepress/examples-src/README.md`
  - Check how examples are generated.
- `docs-vitepress/scripts/`
  - Follow the style of existing docs scripts.

Suggested approach:

- Add a small smoke-test entry point for docs examples.
- Start with only 2 or 3 low-cost examples.
- Avoid heavyweight optional dependency examples such as XGBoost, LightGBM, and CatBoost for the first version.
- Use deterministic seeds and very small `population_size` / `generations` settings where examples run optimization.
- The smoke test should fail with a clear message showing which example failed.

Examples that may be good starting candidates:

- basic usage
- feature selection tutorial
- plotting gallery setup, if it can run cheaply

The final selection can be adjusted after checking runtime.

## Tests to add

Add or document validation for:

1. The smoke-test command runs successfully for the selected examples.
2. A failing example reports the example name or path.
3. Optional dependency examples are skipped or excluded intentionally.

## Validation

Run the new smoke command, for example:

```bash
python docs-vitepress/scripts/<new-smoke-script>.py
```

Also run any focused tests added for helper code.

## Notes for contributors

This is an advanced issue because it touches docs tooling and example runtime behavior. Keep the first version intentionally small so it adds value without making CI slow.

GauravPatil2515:
Resolves #262 by introducing a lightweight smoke test for documentation examples that caps generations and population size using a monkeypatch, and includes pytest validation for this runner.

### DIFF
--- a/docs-vitepress/scripts/smoke_test_examples.py
+++ b/docs-vitepress/scripts/smoke_test_examples.py
@@ -1,0 +1,58 @@
+import os
+import sys
+from pathlib import Path
+import runpy
+
+ROOT = Path(__file__).resolve().parents[2]
+EXAMPLES_SRC = ROOT / "docs-vitepress" / "examples-src"
+
+
+def patch_evolution_config():
+    from sklearn_genetic import EvolutionConfig
+
+    original_init = EvolutionConfig.__init__
+
+    def mocked_init(self, *args, **kwargs):
+        kwargs["population_size"] = min(kwargs.get("population_size", 2), 2)
+        kwargs["generations"] = min(kwargs.get("generations", 1), 1)
+        if "keep_top_k" in kwargs:
+            kwargs["keep_top_k"] = min(kwargs.get("keep_top_k", 1), 1)
+        original_init(self, *args, **kwargs)
+
+    EvolutionConfig.__init__ = mocked_init
+
+
+def run_smoke_tests(target_examples, base_dir=EXAMPLES_SRC):
+    patch_evolution_config()
+    sys.path.insert(0, str(ROOT))
+    sys.path.insert(0, str(base_dir))
+    os.environ["MPLBACKEND"] = "Agg"
+
+    failures = []
+    for filename in target_examples:
+        path = base_dir / filename
+        print(f"-> smoke testing {filename}")
+        try:
+            runpy.run_path(str(path), run_name="__docs_build__")
+        except Exception as exc:
+            print(f"   FAILED: {filename} - {type(exc).__name__}: {exc}")
+            failures.append((filename, exc))
+    return failures
+
+
+def main():
+    target_examples = [
+        "page_basic_usage.py",
+        "page_tutorial_feature_selection.py",
+        "page_plotting_gallery.py",
+    ]
+    failures = run_smoke_tests(target_examples)
+    if failures:
+        print(f"\n{len(failures)} example(s) failed smoke test.")
+        sys.exit(1)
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Hi @GauravPatil2515, thank you for working on this. The PR is aimed at the right issue, and #262 is still open, so this is not a duplicate.

I think this needs a few changes before we can merge it:

- The selected files in `examples-src` are documentation generators, not plain example scripts. Running them with `runpy.run_path` writes generated Markdown and figures into the docs folders, which can leave changes in the working tree during tests/CI.
- `patch_evolution_config()` modifies `EvolutionConfig.__init__` globally and does not restore it afterwards. That could affect other tests running in the same Python process.
- The current patch only caps values passed through `EvolutionConfig(...)`. Some selected examples also pass `population_size` and `generations` directly to estimators, so those runs may still be heavier than intended.
- Since `docs-vitepress/scripts/build_docs.py` already has `discover()` / `run_page()` and a `--check` mode, could you please reuse or extend that existing docs runner instead of adding a separate execution path?

Could you update the implementation so the smoke test has no persistent file side effects, restores any monkeypatching after use, and reliably keeps the selected examples lightweight?

Thanks again!

mayoka0 (commented):
Thoughtful approach. Two things worth flagging before this lands:

1. **Overlap with `build_docs.py --check`** — that script already runs every `page_*.py` generator end to end as a smoke test (that's its `--check` mode). This PR is a lighter, faster subset with shrunk configs, which is a reasonable *complement*, but it'd be worth confirming with the maintainer whether they want a second, overlapping runner vs. e.g. a `--fast` flag on the existing one.

2. **The monkeypatch only covers `EvolutionConfig`** — `mocked_init` shrinks `population_size`/`generations` only for examples that construct an `EvolutionConfig`. Examples that pass those directly to `GASearchCV(...)` (or `GAFeatureSelectionCV(...)`) won't be shrunk and will run at full cost, which partly defeats the "lightweight" goal. Patching at the estimator level (or both) would make it robust.

Solid contribution otherwise.

rodrigo-arenas:
Hi @GauravPatil2515, just checking in on this PR.

Thanks again for taking on #262. The general direction still looks useful, but the requested changes above are still needed before we can move this forward:

- avoid persistent docs/generated-file side effects during the smoke test
- restore any monkeypatching after the test run
- make sure the lightweight caps also apply when examples pass `population_size` / `generations` directly
- preferably reuse or extend the existing `docs-vitepress/scripts/build_docs.py --check` path instead of adding a separate runner

No rush, but if you are still interested in continuing this, an update along those lines would be very welcome. If you are no longer available, no worries either; just let us know so someone else can pick it up.

rodrigo-arenas:
Hi, I'm closing this PR, which doesn't seem to have recent activity, so the issue can be picked again

PR #274, opened 2026-06-28
test(plots): add parametrized tests for _as_list (closes #256) Covers…

KingSylvan:
Adds tests for `_as_list` using pytest.mark.parametrize so each case
runs on its own makes it a lot easier to see exactly what broke.

Covered the cases from the issue and threw in a few extras like empty
string, numpy arrays, and falsy values (0, False) that seemed worth having.

## Summary

Parametrized unit tests for the `_as_list` plotting helper (issue #256).
No behaviour changes — tests only.

## Related Issue

Closes #256

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [x] Other (describe): tests for existing internal helper

## Checklist

- [x] I checked the linked issue is open and not already covered by another PR
- [x] Tests pass locally (`pytest sklearn_genetic/`)
- [ ] Code is formatted with Black (`black sklearn_genetic/`)
- [x] New functionality is covered by tests
- [ ] Documentation updated if needed

PR #270, opened 2026-06-28
test(plots): add direct unit tests for the _as_list helper (#256)

cc1a2b:
## Summary

Add direct unit tests for the internal `_as_list` plotting helper. Test-only — no behavior change.

## Related Issue

Closes #256

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [x] Other (describe): test coverage for an internal helper

## What changed

`_as_list` (in `sklearn_genetic/plots.py`) normalizes the field/parameter-name inputs accepted by several public plotting functions (`plot_history`, `plot_search_space`, `plot_parameter_evolution`, …), where a value may be a single string or an iterable of strings. It had no direct test. This pins its contract:

- `_as_list(None)` → `[]`
- `_as_list("score")` → `["score"]` (a single string is **not** split into characters)
- `_as_list(("a", "b"))` / `_as_list(["a", "b"])` → `["a", "b"]`
- a generator (lazy iterable) is consumed into a concrete list
- a non-iterable scalar is wrapped as `[value]`

## Tests

```
pytest sklearn_genetic/tests/test_plots.py -k as_list
# 1 passed
```

## Checklist

- [x] I checked the linked issue is open and not already covered by another PR
- [x] Tests pass locally (`pytest sklearn_genetic/`)
- [x] Code is formatted with Black (`black sklearn_genetic/`)
- [x] New functionality is covered by tests
- [ ] Documentation updated if needed (not needed — internal helper)

Issue #256, opened 2026-06-28
[Good First Issue] Add tests for empty and string inputs in internal _as_list helper


rodrigo-arenas:
## Summary

Add focused tests for the internal `_as_list` plotting helper.

`_as_list` normalizes plotting inputs such as field names and parameter names. It is small, but it supports several public plotting functions, so a few direct tests would make future refactors safer.

## Why this helps

Several plotting helpers accept either a single string or an iterable of strings. A string should become `["name"]`, not `["n", "a", "m", "e"]`. Tests make that behavior explicit for new contributors.

## Suggested implementation

Start by reading these files:

- `sklearn_genetic/plots.py`
  - Look at `_as_list`.
- `sklearn_genetic/tests/test_plots.py`
  - Add a small test near the other plotting helper tests.

Suggested approach:

- Do not change `_as_list` unless a test exposes an existing bug.
- Add a direct unit test for the helper.
- Keep the test independent from matplotlib where possible.

## Tests to add

Add focused tests in `sklearn_genetic/tests/test_plots.py` for:

1. `_as_list(None)` returns `[]`.
2. `_as_list("score")` returns `["score"]`.
3. `_as_list(("a", "b"))` returns `["a", "b"]`.
4. `_as_list` handles a generator or other iterable by returning a list.

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_plots.py -k as_list
```

## Notes for contributors

This is a good first issue because it is test-focused and helps document a small utility used by the plotting API.

cc1a2b:
Hi! I'd like to take this one. Opening a PR now with direct unit tests for the `_as_list` plotting helper (None/str/iterable/generator/scalar cases). No behavior change.

PR #291, opened 2026-07-01
test(plots): add a few edge cases for _as_list not in #270 empty stri…

KingSylvan:
Follow-up to #270 as suggested by @rodrigo-arenas in #274.

Covers four edge cases that #270 doesn't have- empty string, falsy values
(0, False), numpy arrays, and sets. All felt worth having given how this
helper actually gets called. No behaviour changes, tests only.

## Summary

Adds four edge-case tests for the internal `_as_list` helper that weren't
covered by #270. No behaviour changes, tests only.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [x] Other (describe): additional test coverage for internal helper

## Checklist

### DIFF
--- a/sklearn_genetic/tests/test_plots.py
+++ b/sklearn_genetic/tests/test_plots.py
@@ -1,5 +1,6 @@
 import matplotlib
 import pytest
+import numpy as np
 import re
 from sklearn.datasets import load_diabetes
 from sklearn.model_selection import train_test_split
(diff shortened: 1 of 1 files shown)


rodrigo-arenas:
Hi @KingSylvan, thank you for following up on the suggestion from #274.

This is a good follow-up to #270 and it adds the extra `_as_list` edge cases we discussed: numpy arrays, sets, empty strings, and falsy scalar values. Since #256 was already closed by #270, it is also good that this PR does not try to close the issue again.

I think this looks good overall. One optional cleanup would be to keep the test comments a bit more concise or parametrize the scalar cases, but that is not a blocker from my side.

Thanks again for splitting this into a focused follow-up.

mayoka0 (approved):
Good edge cases — verified each against `_as_list`: numpy array & set hit the `Iterable` branch, `""` correctly returns `[""]` (not `[]`), and `0`/`False` fall through to `[value]`. All correct, and nice call coordinating to not overlap #270.

One heads-up (not a code issue): #270 and #274 also append tests to the end of `test_plots.py`, so whichever lands second will hit a trivial merge conflict at the file tail — easy to resolve, just worth expecting.

rodrigo-arenas (approved):
Thanks @KingSylvan, I re-reviewed this and everything looks good now. The added `_as_list` edge cases are focused, the behavior matches the helper implementation, and the CI checks are passing.

KingSylvan:
Thanks for the review and continued guidance, @rodrigo-arenas  and @mayoka0 much appreciated!

PR #277, opened 2026-06-29
docs(presets): add full Pipeline example for preset prefix (#263)

mayoka0:
## Summary

Adds a compact, runnable example to `api/presets.md` showing how to use a preset search space inside a scikit-learn `Pipeline` — how `prefix="model__"` maps to `model__n_estimators`-style keys and plugs into `GASearchCV(estimator=pipe)`.

## Related Issue

Closes #263

## Type of Change

- [x] Documentation update

## Notes

- Built on the existing "Pipeline prefixes" section (extended, not duplicated).
- Example uses a built-in sklearn dataset (`load_breast_cancer`) and small `population_size`/`generations`, per the docs example standards — verified it constructs and the prefix produces `model__` keys.
- Links to the existing `guide/pipeline-tuning` page (target verified).
- Edited `versions/latest/` only.

## Checklist

- [x] Tests pass locally — n/a (docs)
- [x] Code is formatted with Black — n/a (docs)
- [x] Documentation updated

— Contributed by **Mayoka Labs**

PR #292, opened 2026-07-01
docs: add Using Preset Search Spaces section to pipeline-tuning guide

andrianbalanesq:
## What

Adds a new **Using Preset Search Spaces** section to the pipeline-tuning guide, showing how to use `sklearn_genetic.presets` with `Pipeline` objects.

## Why

Following up on the maintainer invitation from #282 — the preset prefix usage was the missing piece. Presets use bare parameter names, but pipelines need the `step__param` convention. This section shows the dict-comprehension pattern to bridge that gap.

## What changed

- Added **Using Preset Search Spaces** section to `docs-vitepress/versions/latest/guide/pipeline-tuning.md`
- Includes a table of available presets
- Full working example using `GRADIENT_BOOSTING_REG` with a `Pipeline`
- Tip callout for mixing preset + custom parameters

Happy to adjust anything! ✨

### DIFF
--- a/docs-vitepress/versions/latest/guide/pipeline-tuning.md
+++ b/docs-vitepress/versions/latest/guide/pipeline-tuning.md
@@ -98,35 +98,34 @@ y_pred = search.predict(X_test)
 
 ## Using Preset Search Spaces
 
-Manually defining `param_grid` for every estimator gets repetitive. `sklearn_genetic.presets` ships pre-built parameter grids for common models — pass them directly to `GASearchCV` to skip the boilerplate.
+Manually defining `param_grid` for every estimator gets repetitive. `sklearn_genetic` ships pre-built parameter grids for common models — call a preset function and pass the result directly to `GASearchCV` to skip the boilerplate.
 
 ### Available Presets
 
-| Preset | Estimator | Type |
+Each preset is a factory function that returns a ready-to-use `param_grid` dict. Pass `profile="fast"`, `"balanced"` (default), or `"wide"` to control the search range.
+
+| Preset function | Estimator | Type |
 |---|---|---|
-| `GRADIENT_BOOSTING_REG` / `GRADIENT_BOOSTING_CLF` | `GradientBoostingRegressor` / `GradientBoostingClassifier` | Regression / Classification |
-| `RANDOM_FOREST_REG` / `RANDOM_FOREST_CLF` | `RandomForestRegressor` / `RandomForestClassifier` | Regression / Classification |
-| `SVR` / `SVC` | `SVR` / `SVC` | Regression / Classification |
-| `LOGISTIC_REGRESSION` | `LogisticRegression` | Classification |
-| `RIDGE` / `LASSO` / `ELASTIC_NET` | `Ridge` / `Lasso` / `ElasticNet` | Regression |
-| `KNN_REGRESSOR` / `KNN_CLASSIFIER` | `KNeighborsRegressor` / `KNeighborsClassifier` | Regression / Classification |
-| `DECISION_TREE_REG` / `DECISION_TREE_CLF` | `DecisionTreeRegressor` / `DecisionTreeClassifier` | Regression / Classification |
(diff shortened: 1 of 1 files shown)


rodrigo-arenas:
Hi @andrianbalanesq, thank you for opening this focused follow-up from #282.

The placement is right: `guide/pipeline-tuning.md` is the correct page for this section, and this is not a duplicate of #277 because #277 documents the API reference side while this PR adds the guide-side explanation.

However, the current example/table uses preset names that do not exist in the current API, such as `GRADIENT_BOOSTING_REG`, `GRADIENT_BOOSTING_CLF`, `RANDOM_FOREST_REG`, and `SVC`. The current preset API exposes factory functions instead, for example:

- `random_forest_classifier_space`
- `random_forest_regressor_space`
- `hist_gradient_boosting_classifier_space`
- `hist_gradient_boosting_regressor_space`
- `logistic_regression_space`
- `svc_space`
- `xgboost_classifier_space`
- `xgboost_regressor_space`

Could you please update the section to use the actual preset functions? For the pipeline example, it would be better to use the built-in `prefix` argument directly, e.g.:

```python
from sklearn_genetic import random_forest_regressor_space

param_grid = random_forest_regressor_space(prefix="regressor__")

**Review by mayoka0 — COMMENTED** [review]
Thanks for adding this — but heads-up, the preset API used here doesn't match the library, so the example won't run as written:

- The presets are **functions**, not uppercase constants. There's no `GRADIENT_BOOSTING_REG`, `SVR`, `RIDGE`, `LASSO`, `ELASTIC_NET`, `KNN_*`, or `DECISION_TREE_*`. `from sklearn_genetic.presets import GRADIENT_BOOSTING_REG` raises `ImportError`, and `.items()` on it doesn't apply.
- The actual presets (8 of them) are: `random_forest_classifier_space`, `random_forest_regressor_space`, `hist_gradient_boosting_classifier_space`, `hist_gradient_boosting_regressor_space`, `logistic_regression_space`, `svc_space`, `xgboost_classifier_space`, `xgboost_regressor_space`. Each returns a ready `param_grid` and takes a `prefix=` argument, so the pipeline case is simply:

```python
from sklearn_genetic.presets import random_forest_regressor_space
param_grid = random_forest_regressor_space(prefix="regressor__")
```

So the "Available Presets" table and the code block both need to be reworked against the real functions. (FWIW the merged `api/presets.md` "Pipeline prefixes" section has a runnable version you can mirror.) Happy to help if useful.

andrianbalanesq:
Thanks for the thorough review @rodrigo-arenas! Really appreciate you pointing out the correct API (´• ω •`)

Updated both issues:

1. Replaced all the non-existent constant names with the actual factory functions — `random_forest_regressor_space`, `hist_gradient_boosting_classifier_space`, `svc_space`, etc. The table now matches what's exported from `sklearn_genetic`.
2. Switched the pipeline example to use the built-in `prefix` argument directly (`random_forest_regressor_space(prefix="regressor__")`) instead of the manual dict comprehension. Much cleaner — and it's consistent with the example in `api/presets.md`.

Also swapped the example estimator from `GradientBoostingRegressor` (which has no preset) to `RandomForestRegressor` so the snippet actually runs against a real preset.

Pushed to the same branch. Let me know if anything else needs tweaking!

andrianbalanesq:
Thanks for flagging this @mayoka0! You're right — the original version used names that don't exist in the library. I've updated the section to use the actual factory functions (random_forest_regressor_space, hist_gradient_boosting_classifier_space, svc_space, etc.) and switched the pipeline example to use the built-in prefix argument directly, mirroring the approach in api/presets.md. Also swapped the estimator to RandomForestRegressor so the snippet runs against a real preset. Pushed to the same branch — let me know if anything else needs adjusting! ✨

rodrigo-arenas (approved):
Thanks @andrianbalanesq, I re-reviewed the latest update and this looks good now. The section now uses the actual preset factory functions, the `prefix` argument is the right pipeline approach, and the checks are passing.

PR #303, opened 2026-07-03
Fix duplicate param_grid key in ModelCheckpoint.estimator_state and add regression test for core state keys

### LINKED ISSUE #294: [Good first issue] Add a small regression test for ModelCheckpoint state keys
## Summary

`ModelCheckpoint` builds its `estimator_state` dictionary manually and currently
includes `param_grid` twice. Clean up the duplicate key and add a focused
regression test for the expected checkpoint state keys.

## Why this helps

Checkpoint files are used to resume or inspect long-running searches. Keeping
their structure explicit and covered by tests makes future callback changes less
fragile.

## Suggested implementation

Start by reading:

  - Inspect the `estimator_state` dictionary in `ModelCheckpoint.on_step`.
  - Look at the existing `ModelCheckpoint` coverage.

Remove the duplicate `param_grid` entry and add a test that saves a checkpoint
and verifies the important keys in `checkpoint_data["estimator_state"]`.

## Tests to add

Add or extend tests in `sklearn_genetic/tests/test_persistence_and_helpers.py`
for:

1. a checkpoint contains expected core keys such as `estimator`, `cv`, `scoring`,
   `population_size`, `generations`, `param_grid`, and `algorithm`
2. the saved checkpoint can still be loaded through `ModelCheckpoint.load()`

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_persistence_and_helpers.py
```

## Notes for contributors

This is a good first issue because the change is localized and the existing test
file already has a nearby pattern to follow.

AndyDLi:
## Summary

`ModelCheckpoint.on_step` built its `estimator_state` dictionary with the `param_grid` key assigned twice, silently overwriting the first entry. This PR removes the duplicate and adds a regression test covering the checkpoint's core state keys.

## Related Issue

Closes #294

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] Other (describe):

## Checklist

### DIFF
--- a/sklearn_genetic/callbacks/model_checkpoint.py
+++ b/sklearn_genetic/callbacks/model_checkpoint.py
@@ -25,7 +25,6 @@ class ModelCheckpoint(BaseCallback):
                 "mutation_probability": estimator.mutation_probability,
                 "param_grid": estimator.param_grid,
                 "algorithm": estimator.algorithm,
-                "param_grid": estimator.param_grid,
                 "local_search": getattr(estimator, "local_search", False),
                 "local_search_top_k": getattr(estimator, "local_search_top_k", 1),
                 "local_search_steps": getattr(estimator, "local_search_steps", 1),
--- a/sklearn_genetic/tests/test_persistence_and_helpers.py
+++ b/sklearn_genetic/tests/test_persistence_and_helpers.py
@@ -234,5 +234,7 @@ def test_model_checkpoint_estimator_state_keys(tmp_path):
     assert estimator_state["scoring"] == "accuracy"
     assert estimator_state["population_size"] == 4
     assert estimator_state["generations"] == 2
-    assert estimator_state["algorithm"] == "eaSimple"
     assert list(estimator_state["param_grid"].keys()) == ["max_depth"]
+    assert estimator_state["param_grid"]["max_depth"].lower == 1
+    assert estimator_state["param_grid"]["max_depth"].upper == 2
+    assert estimator_state["algorithm"] == "eaSimple"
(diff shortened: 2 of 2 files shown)


mayoka0 (approved):
Verified this for real, not just by reading the diff.

- Confirmed the duplicate on `master`: `"param_grid": estimator.param_grid,` appears twice in the `estimator_state` dict literal (once after `algorithm`, once after `mutation_probability`). Checked it's a true duplicate, not a copy-paste that accidentally clobbered a *different* intended key — every other field (`local_search`, `diversity_control`, etc.) is present and distinct, so nothing else is silently missing.
- Ran the full test file on this branch: all 10 tests pass, including the new one.
- Checked the actual runtime effect: since both duplicate assignments hold the identical value, Python's dict-literal semantics (last value wins, but the *key's position* is set by the first occurrence) mean removing the duplicate produces a byte-identical `estimator_state` — so this is safe dead-code removal, not a behavior change.

One minor, non-blocking observation: because both duplicate values were identical, `test_model_checkpoint_estimator_state_keys` wouldn't actually *fail* if the duplicate line were reintroduced (the resulting dict would be unchanged either way) — so it's a good general coverage test for the checkpoint's core keys, just not a tight regression pin for this specific bug. Not asking for a change, just flagging it.

Clean, verified fix. LGTM.

rodrigo-arenas (approved):
Thanks @AndyDLi, I re-reviewed this and it looks good. The duplicate `param_grid` entry is removed, the checkpoint still saves and loads through the existing path, and the new test covers the expected core state keys from #294.

Issue #293, opened 2026-07-02
[Good first issue] Fix outdated external references link in README and CONTRIBUTING


rodrigo-arenas:
## Summary

The contributor-facing external references link is stale. `CONTRIBUTING.md` and
`README.rst` still point to `docs/external_references.rst`, but that file is no
longer present in the repository.

## Why this helps

New contributors may follow this link when they want to add articles, videos, or
other community references. A broken or outdated path makes a first contribution
more confusing than it needs to be.

## Suggested implementation

Start by reading:

- `CONTRIBUTING.md`
  - Look for the "External References" section.
- `README.rst`
  - Look for the community/reference contribution text near the end.
- `docs-vitepress/versions/latest/community/articles.md`
  - Confirm whether this is now the canonical place for community links.

Update the outdated references to point to the current docs page or adjust the
wording so contributors know where to add community articles today.

## Tests to add

No Python tests are required. If the docs link checker is extended in another
issue, this should also pass that check.

## Validation

Run:

```bash
rg "external_references|docs/external" README.rst CONTRIBUTING.md docs-vitepress
```

The old `docs/external_references.rst` target should no longer appear in
contributor-facing text.

## Notes for contributors

This is a good first issue because it is documentation-only, small, and easy to
verify locally.

qorexdevs:
opened #302 for this - swapped the dead docs/external_references.rst link for the Community Articles page in both files

PR #305, opened 2026-07-03
docs: check root doc links

### LINKED ISSUE #298: [Intermediate] Add a root docs link checker for README and CONTRIBUTING
## Summary

Extend documentation link checking so stale links in root docs such as
`README.rst` and `CONTRIBUTING.md` are detected.

## Why this helps

The existing VitePress link checker covers versioned Markdown docs, but it does
not catch stale links in top-level contributor files. For example, the old
`docs/external_references.rst` link survived there after the docs structure
changed.

## Suggested implementation

Start by reading:

- `docs-vitepress/scripts/check_links.py`
  - Understand the current relative Markdown link resolver.
- `docs-vitepress/scripts/test_check_links.py`
  - Follow the existing testing style.
- `README.rst` and `CONTRIBUTING.md`
  - Identify the kinds of links that should be checked.

Either extend the existing script or add a small companion checker for root
`.rst`/`.md` files. Keep the scope practical: checking local repository paths is
enough; external HTTP link validation is not required.

## Tests to add

Add tests for:

1. a valid local root-doc link passes
2. a missing local file path is reported
3. external URLs are skipped

## Validation

Run:

```bash
pytest docs-vitepress/scripts/test_check_links.py
python docs-vitepress/scripts/check_links.py
```

## Notes for contributors

This is an intermediate issue because it involves documentation tooling and
parsing, but it should still be well-contained.

vku2018:
## Summary
- extend `docs-vitepress/scripts/check_links.py` to scan root `README.rst` and `CONTRIBUTING.md` in addition to versioned VitePress Markdown docs
- add RST target extraction for inline links, reference definitions, image directives, and directive targets
- add tests for valid root-doc links, missing local root-doc links, external link skipping, and an RST prose false-positive guard

## Validation
- `python -m pytest docs-vitepress/scripts/test_check_links.py`
- `PYTHONIOENCODING=utf-8 python docs-vitepress/scripts/check_links.py`
- `.venv/Scripts/python.exe -m black --check docs-vitepress/scripts/check_links.py docs-vitepress/scripts/test_check_links.py`
- `git diff --check`

Note: only Python 3.11.9 is available in this environment. Black reported both touched files unchanged, with its standard warning that the local Python is older than the project target.

Closes #298

### DIFF
--- a/docs-vitepress/scripts/check_links.py
+++ b/docs-vitepress/scripts/check_links.py
@@ -1,9 +1,9 @@
-"""Check internal Markdown links in the versioned VitePress docs.
+"""Check internal documentation links.
 
-Scans every Markdown file under the versioned docs trees and verifies that
-*relative* links point at a file that exists. VitePress lets you omit the
-``.md`` extension and link to a directory's ``index.md``, so this resolver
-mirrors that behaviour.
+Scans every Markdown file under the versioned docs trees, plus the root
+contributor docs, and verifies that *relative* links point at a file that
+exists. VitePress lets you omit the ``.md`` extension and link to a directory's
+``index.md``, so this resolver mirrors that behaviour.
 
 What is checked / skipped:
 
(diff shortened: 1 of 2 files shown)


mayoka0 (approved):
Nice extension — ran it against the real repo, not just the tests.

- All 15 tests pass, and running `check_links.py` against the actual `README.rst`/`CONTRIBUTING.md` on this branch is clean (no false positives from the new RST regexes).
- Checked the badge/image usage in `README.rst` (`.. image::` + `:target:` lines) — all external `https://` URLs, correctly filtered out by the existing `_is_internal_relative`, so no noise there.

One thing worth knowing (not a defect, just sets expectations): the actual broken reference that #293/#301/#302/#304 are fixing — `docs/external_references.rst` — is linked via a **full `https://github.com/.../blob/master/docs/external_references.rst` URL**, not a relative path. Since this checker is scoped to *relative* links by design, it would **not** have caught that specific bug even after merging. Might be worth a one-line README/docstring note so nobody assumes this PR prevents a recurrence of the exact issue that prompted it — the actual guard against that class of breakage would need to check external self-referential GitHub links, which is a different (and heavier) tool.

Also minor/non-blocking: `RST_DIRECTIVE_TARGET_RE` only matches `.. image::`, not `.. figure::` (also a standard RST image directive). Not currently used anywhere in this repo's docs, so no active impact — just a gap if someone adds a `.. figure::` later.

Solid work either way, and good tests (including the RST-prose-with-angle-brackets false-positive guard, which is exactly the kind of case regex-based RST parsing tends to trip on). LGTM.

rodrigo-arenas:
Thanks for working on this! This PR now conflicts with `master`, and the same issue was already addressed by #321, which has been merged.

Since #321 covers the root README/CONTRIBUTING link checking scope for #298, I’m going to close this PR as superseded. Appreciate the contribution!

PR #310, opened 2026-07-03
Restore fitness cache on checkpoint resume (part of #299)

### LINKED ISSUE #299: [Advanced] Make checkpoint resume preserve full optimization runtime state
## Summary

Audit and improve checkpoint resume so it preserves the full optimization
runtime state needed for reliable continuation.

## Why this helps

Long-running searches need dependable resume behavior. Today the fit flow loads
`estimator_state` and `logbook` from `ModelCheckpoint`, but it is worth auditing
whether cache state, history, generation progress, random number state, and
`fit_stats_` are restored completely and consistently.

## Suggested implementation

Start by reading:

  - Review the checkpoint loading paths in both `GASearchCV.fit` and
    `GAFeatureSelectionCV.fit`.
  - Review what is saved today.
- `sklearn_genetic/_base.py`
  - Review the generic `save`/`load` behavior.
  - Review existing checkpoint tests.

Design a resume test that intentionally stops or checkpoints a short search,
resumes it, and verifies that already evaluated candidates are not needlessly
re-evaluated and that final results are reproducible when `random_state` is set.

## Tests to add

Add tests for:

1. resumed runs preserve or reconstruct `history`, `logbook`, and `fit_stats_`
2. cached candidates are not re-evaluated after resume when `use_cache=True`
3. `random_state` plus checkpoint resume produces deterministic results for a
   small search

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_genetic_search.py sklearn_genetic/tests/test_persistence_and_helpers.py
```

## Notes for contributors

This is an advanced issue because it touches fit-loop state, serialization, and
reproducibility. Please discuss the intended design before opening a large PR.

mayoka0:
Delivers the **cache-reuse** portion of #299. **This does not close #299** — the broader resume-state work is intentionally left open (see scope note).

## Problem
When a run resumes from a `ModelCheckpoint`, `fit()` restores the estimator config and `logbook`, but **not** the `fitness_cache`. Every candidate evaluated before the checkpoint is therefore re-evaluated on resume.

## Change
- `ModelCheckpoint.on_step` saves the estimator's `fitness_cache` under a new, separate `runtime_state` key.
- `GASearchCV.fit` and `GAFeatureSelectionCV.fit` restore it on resume.

`runtime_state` is kept **separate** from `estimator_state` because the latter is consumed as constructor kwargs (`GASearchCV(**checkpoint_data["estimator_state"])`); adding non-constructor keys there would break that contract. The restore is guarded with `.get("runtime_state")`, so **pre-#299 checkpoint files still load** unchanged.

## Tests (addressing the review)
- `test_checkpoint_resume_restores_fitness_cache` — GASearchCV restore branch (sentinel cache entry survives resume).
- `test_feature_selection_checkpoint_resume_restores_fitness_cache` — **GAFeatureSelectionCV restore branch** (the PR touches both fit methods).
- `test_checkpoint_resume_without_runtime_state_is_backward_compatible` — **backward-compat**: a checkpoint with no `runtime_state` resumes cleanly.

These exercise both restore branches and the `.get()` fallback, covering the previously-missing `codecov/patch` lines. `black --check` clean.

## Scope note (re: your review)
You're right that #299 is broader than the cache. In particular, as you noted, `_register()` runs after checkpoint loading and resets `self.logbook`, so logbook/history state is not truly preserved yet. Rather than overreach, this PR is scoped to the cache-reuse win and **does not close #299**; the remaining runtime-state work (history/logbook/fit_stats_ preservation, deterministic resume) is left open for a focused follow-up.

### DIFF
--- a/sklearn_genetic/callbacks/model_checkpoint.py
+++ b/sklearn_genetic/callbacks/model_checkpoint.py
@@ -40,7 +40,16 @@ class ModelCheckpoint(BaseCallback):
                 "sharing_radius": getattr(estimator, "sharing_radius", 0.2),
                 "sharing_alpha": getattr(estimator, "sharing_alpha", 1.0),
             }
-            checkpoint_data = {"estimator_state": estimator_state, "logbook": deepcopy(logbook)}
+            # Runtime state is kept separate from ``estimator_state`` so the
+            # latter stays constructor-compatible (it is consumed as
+            # ``GASearchCV(**estimator_state)``). Restoring the fitness cache on
+            # resume lets already-evaluated candidates be reused (see fit()).
+            runtime_state = {"fitness_cache": getattr(estimator, "fitness_cache", {})}
+            checkpoint_data = {
+                "estimator_state": estimator_state,
+                "logbook": deepcopy(logbook),
+                "runtime_state": runtime_state,
+            }
             with open(self.checkpoint_path, "wb") as f:
                 pickle.dump(checkpoint_data, f)
                 print(f"Checkpoint saved to {self.checkpoint_path}")
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -1073,6 +1073,14 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
                     if checkpoint_data:
                         self.__dict__.update(checkpoint_data["estimator_state"])  # noqa
                         self.logbook = checkpoint_data["logbook"]
+                        # Restore the fitness cache so already-evaluated
+                        # candidates are reused instead of re-evaluated. Older
+                        # checkpoints have no ``runtime_state``, so guard with
+                        # .get() for backward compatibility.
+                        runtime_state = checkpoint_data.get("runtime_state") or {}
+                        cached = runtime_state.get("fitness_cache")
+                        if cached is not None:
+                            self.fitness_cache = cached
                         checkpoint_loaded = True
                     break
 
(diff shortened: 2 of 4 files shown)


rodrigo-arenas:
Hi @mayoka0, thanks for tackling this. Restoring `fitness_cache` under a separate `runtime_state` key is a sensible direction, and keeping it out of `estimator_state` preserves the constructor-compatible checkpoint contract.

I do not think this is ready to merge yet, mainly because #299 is broader than the cache alone. The issue asks for resume coverage around `history`, `logbook`, `fit_stats_`, cache reuse, and deterministic resume behavior. This PR currently closes #299 but only verifies the cache path for `GASearchCV`.

Could you please adjust this before merge?

- Either expand the implementation/tests to cover the full #299 resume-state scope, or change the PR so it does not close #299 and clearly leaves the remaining runtime-state work open.
- Add coverage for the `GAFeatureSelectionCV` restore path too, since the PR changes both fit methods.
- Add a backward-compatibility test for loading an older checkpoint without `runtime_state`.
- Address the failing `codecov/patch` check; the missing lines are in the new restore branches.

One extra thing I noticed while reviewing the current fit flow: `_register()` runs after checkpoint loading and resets `self.logbook`, so the existing resume path still does not truly preserve logbook/history state. That reinforces why I would avoid closing #299 with only the cache portion.

Thanks again; this is a useful improvement, it just needs the scope and tests tightened before merge.

mayoka0:
Apologies for the slow turnaround, @rodrigo-arenas — I was away for a couple of days and did not get back to your review as promptly as I should have. Thanks for the careful, detailed feedback; it genuinely made these better.

Thanks for the detailed review, @rodrigo-arenas. Updated to address each point:

- **Scope / #299 closure.** Agreed the issue is broader than the cache. This PR no longer closes #299 — it's scoped to the cache-reuse portion, and I've left the runtime-state work (history/logbook/`fit_stats_`, deterministic resume) open. Your observation that `_register()` resets `self.logbook` after load is exactly why I didn't want to claim logbook/history preservation here.
- **GAFeatureSelectionCV restore coverage.** Added `test_feature_selection_checkpoint_resume_restores_fitness_cache`, which drives the FS restore branch directly.
- **Backward-compatibility.** Added `test_checkpoint_resume_without_runtime_state_is_backward_compatible` — a pre-#299 checkpoint (no `runtime_state`) resumes without error.
- **codecov/patch.** The three tests now exercise both restore branches and the `.get()` fallback, so the previously-missing lines are covered.

Rebased onto current `master`. Happy to keep iterating.

rodrigo-arenas:
Thanks for the update. The scoped cache-resume change looks good to me now: `runtime_state` keeps `fitness_cache` out of constructor-compatible `estimator_state`, both restore branches are covered, and the backward-compat path for older checkpoints is tested.

I verified locally:
- targeted new tests -> 3 passed

Thanks again!

Issue #300, opened 2026-07-02
[Advanced] Benchmark and optimize duplicate candidate evaluation/cache behavior


rodrigo-arenas:
## Summary

Benchmark and optimize duplicate candidate evaluation and cache behavior.

## Why this helps

The evaluation layer already deduplicates candidates and tracks cache hits, but
spaces with many collisions can still be performance-sensitive. Better
benchmarks will make future cache or evaluation changes measurable instead of
guesswork.

## Suggested implementation

Start by reading:

- `sklearn_genetic/evaluation.py`
  - Review `evaluate_population`, cache hits, duplicate handling, and
    `fit_stats_` counters.
- `benchmarks/benchmark_fit.py`
  - Follow the existing benchmark style and CLI conventions.
- `sklearn_genetic/tests/test_genetic_search.py`
  - Review existing tests for cache and duplicate behavior.

Add a benchmark scenario with a deliberately collision-prone search space, such
as small integer/categorical domains with population sizes larger than the
number of unique combinations. Compare `use_cache=True` and `use_cache=False`
and report counters like `cache_hits`, `duplicate_candidates`, and
`cross_validate_calls`.

## Tests to add

Add focused tests only if implementation behavior changes. If this is benchmark
only, make sure the benchmark has a quick mode suitable for local smoke checks.

## Validation

Run:

```bash
python benchmarks/benchmark_fit.py --quick
pytest sklearn_genetic/tests/test_genetic_search.py
```

## Notes for contributors

This is an advanced issue because performance work needs careful measurement and
must not change `cv_results_`, logbook contents, or public estimator behavior by
accident.

ShiHuiwen-creat:
Hi @rodrigo-arenas, I'd like to work on the remaining benchmark slice of this issue. I see #311 already added the cache toggle, so to avoid overlap I'll keep the PR focused on a deliberately collision-prone scenario in `benchmarks/benchmark_fit.py`: small Integer/Categorical domains, cache enabled versus disabled, and the existing `cache_hits` / `duplicate_candidates` / `cross_validate_calls` counters. I'll preserve a fast `--quick` path and won't change production cache behavior. I'll add focused coverage for any benchmark helper logic introduced. Please let me know if you prefer a different remaining slice.

PR #311, opened 2026-07-03
Make fitness cache toggleable in benchmark harness (#300)

mayoka0:
Addresses #300 (making cache/duplicate-evaluation behavior measurable).

## Problem
`benchmarks/benchmark_fit.py` reports `actual_cross_validate_calls` and `duplicate_or_cache_reuses`, but `use_cache=True` was **hardcoded** in both `build_gasearch` and `build_feature_selector`. There was no way to run the harness with the cache off, so you couldn't actually measure what the fitness cache buys you.

## Change
- Thread a `use_cache: bool = True` parameter through `build_gasearch` and `build_feature_selector` (replacing the hardcoded `True`).
- Expose it on the CLI as `--use-cache` / `--no-use-cache` (`argparse.BooleanOptionalAction`, default on — existing behavior unchanged).

## Verified locally
Same scenario, quick mode, cache on vs off:

```
--use-cache     cross_validate_calls=25   cache_reuses=1
--no-use-cache  cross_validate_calls=26   cache_reuses=0
```

With `--no-use-cache` every candidate is re-evaluated (zero reuses, more `cross_validate` calls); on a full-size run the gap is far larger. The feature-selection path was smoke-tested the same way. `black --check` is clean. Default runs are unchanged, so existing baselines stay comparable.

### DIFF
--- a/benchmarks/benchmark_fit.py
+++ b/benchmarks/benchmark_fit.py
@@ -259,6 +259,7 @@ def build_gasearch(
     n_jobs: int | None,
     parallel_backend: str,
     population_initializer: str,
+    use_cache: bool = True,
     final_selection: bool = False,
     final_selection_top_k: int = 3,
     final_selection_cv=None,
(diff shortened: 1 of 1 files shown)


rodrigo-arenas (approved):
Thanks @mayoka0, I re-reviewed this and it looks good. The benchmark keeps cache enabled by default, now exposes `--use-cache` / `--no-use-cache`, and threads the setting through both search builders so cache impact can be measured directly.

PR #308, opened 2026-07-03
feat(callbacks): clearer ModelCheckpoint save message (#295)

mayoka0:
## Summary

Polishes the `ModelCheckpoint` success message: `Checkpoint save in {path}` → `Checkpoint saved to {path}` (clearer, grammatical).

## Related Issue

Closes #295

## What changed
- `sklearn_genetic/callbacks/model_checkpoint.py` — the success `print`.
- `test_persistence_and_helpers.py` — updated the `capsys` assertion.
- `docs-vitepress/versions/latest/examples/checkpointing.md` — the captured output (7 lines).

Left `versions/0.13/` untouched (released/frozen per the docs versioning policy), and left the error messages (`Error saving checkpoint` / `Error loading checkpoint`) as-is since they're already clear.

## Checklist
- [x] Tests pass (`pytest sklearn_genetic/tests/test_persistence_and_helpers.py`) — 9 passed
- [x] Black clean
- [x] Docs kept in sync

— Contributed by **Mayoka Labs**

PR #313, opened 2026-07-03
Fix GAFeatureSelectionCV checkpointing via shared state builder (#297)

### LINKED ISSUE #297: [Intermediate] Unify checkpoint serialization between ModelCheckpoint and save/load
## Summary

`GeneticEstimatorMixin.save`.

## Why this helps

## Suggested implementation

Start by reading:

- `sklearn_genetic/_base.py`

## Tests to add

Add tests for:

## Validation

Run:

```bash
```

## Notes for contributors

mayoka0:
Fixes a real, silent bug in `ModelCheckpoint` and centralizes the checkpoint state construction. **This does not close #297** — see scope note below.

## Bug: feature-selection checkpointing is silently broken
`ModelCheckpoint.on_step` built `estimator_state` as a hand-written literal that read `estimator.param_grid` **unconditionally**. `GAFeatureSelectionCV` has no `param_grid`, so every `on_step` raised `AttributeError`, which the surrounding `try/except` swallowed as `Error saving checkpoint: ...`. The estimator state was therefore **never saved for feature selectors**, so checkpoint/resume silently did nothing.

## Change
Extracted a single `_build_estimator_state()` helper backed by one declared set of constructor-compatible keys, replacing the literal. `param_grid` is only included when the estimator defines it, so both `GASearchCV` and `GAFeatureSelectionCV` serialize correctly, and the key set can no longer drift or duplicate.

`estimator_state` stays constructor-compatible — the `GASearchCV` key set is unchanged, so `GASearchCV(**checkpoint_data["estimator_state"])` and the resume path keep working.

## Test
`test_model_checkpoint_supports_feature_selection_estimator`: fits a `GAFeatureSelectionCV` with a `ModelCheckpoint`, asserts no "Error saving checkpoint", that the reloaded state has no `param_grid`, and — per review — that the saved state **round-trips through the resume path to a fitted selector** (`resumed.support_`). Existing checkpoint/persistence tests pass; `black --check` clean; patch coverage is full.

## Scope (re: #297)
Per review, this PR is intentionally narrowed to the concrete feature-selection bug plus centralizing the checkpoint state. It **does not** attempt the broader #297 goal of sharing/validating a contract with `GeneticEstimatorMixin.save/load` (whose `_serializable_state()` uses a different, non-constructor contract). #297 is left open for a focused follow-up.

Rebased onto current `master`, preserving the merged `Checkpoint saved to ...` message (#308) and the dedup from #303.

### DIFF
--- a/sklearn_genetic/callbacks/model_checkpoint.py
+++ b/sklearn_genetic/callbacks/model_checkpoint.py
@@ -3,6 +3,52 @@ from .base import BaseCallback
 from .loggers import LogbookSaver
 from copy import deepcopy
 
+# Single source of truth for the constructor-compatible state persisted in a
+# checkpoint, so the key set cannot silently drift or duplicate.
+#
+# Core keys exist on both GASearchCV and GAFeatureSelectionCV and are read
+# directly. Optional keys map to the default used when the estimator does not
+# define them.
+_CHECKPOINT_CORE_KEYS = (
+    "estimator",
+    "cv",
+    "scoring",
+    "population_size",
+    "generations",
+    "crossover_probability",
+    "mutation_probability",
+    "algorithm",
+)
+_CHECKPOINT_OPTIONAL_DEFAULTS = {
+    "local_search": False,
+    "local_search_top_k": 1,
+    "local_search_steps": 1,
+    "local_search_radius": 0.1,
+    "diversity_control": False,
+    "diversity_threshold": 0.1,
+    "diversity_stagnation_generations": 5,
+    "diversity_mutation_boost": 2.0,
+    "random_immigrants_fraction": 0.1,
+    "fitness_sharing": False,
+    "sharing_radius": 0.2,
+    "sharing_alpha": 1.0,
+}
+
+
+def _build_estimator_state(estimator):
+    """Build the constructor-compatible checkpoint state for an estimator.
+
+    Works for both GASearchCV and GAFeatureSelectionCV: ``param_grid`` is only
+    included when the estimator actually defines it (GAFeatureSelectionCV does
+    not), so checkpointing a feature selector no longer raises AttributeError.
+    """
+    state = {key: getattr(estimator, key) for key in _CHECKPOINT_CORE_KEYS}
+    if hasattr(estimator, "param_grid"):
+        state["param_grid"] = estimator.param_grid
+    for key, default in _CHECKPOINT_OPTIONAL_DEFAULTS.items():
+        state[key] = getattr(estimator, key, default)
+    return state
+
 
 class ModelCheckpoint(BaseCallback):
     def __init__(self, checkpoint_path, **dump_options):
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Hi @mayoka0, thanks for digging into this. The `GAFeatureSelectionCV` checkpoint failure you found is real and worth fixing, and the constructor-compatible state builder is a good direction.

This cannot be merged as-is right now because GitHub reports the branch as conflicting with `master`. #303 and #308 have since landed in `model_checkpoint.py`, so please rebase/update this branch and preserve both of those changes, especially the clearer `Checkpoint saved to ...` message.

I also think the scope needs one more pass before this should close #297. The issue is specifically about unifying `ModelCheckpoint` serialization with `GeneticEstimatorMixin.save/load`; this PR removes the hand-written literal inside `ModelCheckpoint`, but it still does not share a helper or contract with the save/load path. Could you either:

- update the implementation/tests so the checkpoint state and save/load state are intentionally compared or share a helper for the constructor-compatible subset, or
- narrow the PR so it fixes the `GAFeatureSelectionCV` checkpoint bug without closing #297, leaving the broader serialization-unification work open.

It would also help to add a round-trip assertion that the saved `GAFeatureSelectionCV` checkpoint state can actually be used by the resume path, not only that `param_grid` is absent.

Thanks again; this is close in spirit, but it needs the conflict resolved and the #297 contract clarified before merge.

rodrigo-arenas:
Hi @mayoka0, just following up on the earlier review note: this PR is still showing conflicts with \\master\\, so it cannot be merged or fully compared against the other #297 work yet.\n\nAfter reviewing #315 as well, I still think this PR's direction is promising because it preserves the constructor-compatible checkpoint state while addressing the \\GAFeatureSelectionCV\\ checkpoint issue. Could you please rebase/update the branch and keep the changes that have since landed in \\model_checkpoint.py\\?\n\nOnce the conflicts are resolved, the remaining thing to clarify is whether this should fully close #297 by sharing/validating the intended checkpoint-state contract with save/load, or be narrowed to the feature-selection checkpoint bug and leave the broader serialization unification open. Thanks again.

mayoka0:
Apologies for the slow turnaround, @rodrigo-arenas — I was away for a couple of days and did not get back to your review as promptly as I should have. Thanks for the careful, detailed feedback; it genuinely made these better.

Thanks for the review, @rodrigo-arenas. I've updated the PR to address each point:

- **Conflict resolved.** Rebased onto current `master`, preserving both landed changes in `model_checkpoint.py` — the `Checkpoint saved to ...` message (#308) and the dedup (#303). The shared builder also makes that duplicate structurally impossible going forward.
- **Scope narrowed.** This PR no longer closes #297. It fixes the concrete `GAFeatureSelectionCV` checkpoint `AttributeError` and centralizes the checkpoint state, but leaves the broader save/load unification open for a focused follow-up, since `_serializable_state()` intentionally uses a different (non-constructor) contract.
- **Round-trip assertion added.** The regression test now also resumes a fit from the saved feature-selector checkpoint and asserts it produces a fitted selector (`resumed.support_`), not just that `param_grid` is absent.

Patch coverage is full and `black --check` is clean. Happy to adjust further.

rodrigo-arenas:
Hi @mayoka0, thanks for the update. The scoped direction looks good now, and the added round-trip assertion is exactly the right coverage.

GitHub is still reporting this PR as conflicting with current `master`. The new conflict is likely from #310 landing in `model_checkpoint.py`: current `master` now saves `runtime_state = {"fitness_cache": ...}` alongside `estimator_state`. When rebasing, please keep that runtime-state/cache behavior and combine it with this PR's `_build_estimator_state(estimator)` helper, so the final checkpoint data includes both:

- constructor-compatible `estimator_state` from the helper, including `param_grid` only when present;
- `runtime_state["fitness_cache"]` from #310;
- the existing `logbook` and `Checkpoint saved to ...` behavior.

One small admin detail too: GitHub still detects this PR as closing #297, probably because the body says "does not close #297". Could you reword that to avoid the closing keyword, e.g. "Related to #297" / "does not fully resolve issue #297", so #297 remains open after this lands?

After the rebase/conflict fix, this should be ready for another pass.

mayoka0:
Thanks, @rodrigo-arenas — and congrats on getting #310 in.

Rebased onto current `master` and combined the two, as you suggested. `on_step` now:
- keeps #310's `runtime_state = {"fitness_cache": ...}` cache persistence, and
- builds `estimator_state` via the single `_build_estimator_state()` helper (replacing the literal), which only includes `param_grid` when the estimator defines it.

So the feature-selector `AttributeError` is fixed while the cache/runtime-state behavior from #310 is preserved. `GASearchCV`'s key set is unchanged.

The FS round-trip test now uses a **real** saved checkpoint (save works after this fix) and asserts the resumed selector is fitted; it sits alongside your merged `test_feature_selection_checkpoint_resume_restores_fitness_cache`. Full `test_persistence_and_helpers.py` passes (12) and `black --check` is clean.

rodrigo-arenas:
Thanks for the update, @mayoka0. I re-reviewed the latest version and this looks good to merge now.

I verified that the conflict with #310 was resolved correctly: `ModelCheckpoint.on_step` now uses `_build_estimator_state(estimator)` for the constructor-compatible `estimator_state`, while still preserving the `runtime_state["fitness_cache"]` behavior that landed in #310, plus the existing `logbook` and `Checkpoint saved to ...` behavior.

Local verification:
- `test_model_checkpoint_supports_feature_selection_estimator` -> passed
- `test_feature_selection_checkpoint_resume_restores_fitness_cache` -> passed
- `test_model_checkpoint_estimator_state_keys` -> passed

CI is green and the PR is mergeable. The only remaining thing is GitHub still appears to detect #297 as a closing reference because of the wording in the PR body, but I do not want to block this fix on that any longer. If #297 gets closed automatically after merge, we can reopen it since the broader save/load unification work remains open.

PR #315, opened 2026-07-04
Unify ModelCheckpoint and save()/load() checkpoint serialization

### LINKED ISSUE #297: [Intermediate] Unify checkpoint serialization between ModelCheckpoint and save/load

AndyDLi:
## Summary

Unifies checkpoint serialization between `ModelCheckpoint` and `GeneticEstimatorMixin.save()`/`load()`. `ModelCheckpoint.on_step` now reuses `estimator._serializable_state()` when available (falling back to the previous manual dict for non-mixin estimators), instead of maintaining a separately hand-written key list that could drift from what `save()` captures. Also excludes `callbacks` from `_serializable_state()`, since callback objects (including `ModelCheckpoint` itself, mid-`fit()`) shouldn't be pickled as estimator state.

One user-facing behavior change I noticed: `estimator_state` can no longer be unpacked directly as constructor kwargs (`GASearchCV(**estimator_state)`), since it now includes fit-time-only attributes that aren't valid `__init__` arguments. This pattern was never documented anywhere and appeared in exactly one place in the codebase — an internal test (`test_checkpoint_functionality`) — which has been updated to use `.load()` instead, matching the pattern every other save/load test in the suite already used. `.load()` is the actual supported restore mechanism and is unaffected by this change.

## Related Issue

Closes #297

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] Other (describe):

## Checklist

### DIFF
--- a/sklearn_genetic/_base.py
+++ b/sklearn_genetic/_base.py
@@ -17,7 +17,7 @@ def history_record(history, index):
 
 
 class GeneticEstimatorMixin:
-    _volatile_pickle_attrs = {"toolbox", "_stats", "_pop", "_hof", "hof"}
+    _volatile_pickle_attrs = {"toolbox", "_stats", "_pop", "_hof", "hof", "callbacks"}
 
     def _serializable_state(self):
         return {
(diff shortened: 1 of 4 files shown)


rodrigo-arenas:
Thanks for the contribution here. We ended up landing the #297 serialization follow-up through #328, preserving the constructor-compatible checkpoint contract while validating it against save/load, so this PR is now superseded. Closing to keep the queue tidy.

PR #320, opened 2026-07-04
Fix Categorical priors being ignored during sampling

### LINKED ISSUE #314: Categorical priors are ignored during sampling (documented weights have no effect)
### Description
`Categorical(choices=..., priors=...)` validates and stores `priors`, but they are **never used when sampling**, so every choice is drawn with equal probability. This silently contradicts the documented behavior of the `priors` argument:

> priors: List with the probability of sampling each element of the "choices", if not set gives equal probability.

### Reproduction
```python
from collections import Counter
from sklearn_genetic.space import Categorical

dim = Categorical(["a", "b"], priors=[0.98, 0.02], random_state=0)
print(Counter(dim.sample() for _ in range(4000)))
# expected ~ {'a': ~3920, 'b': ~80}
# actual   ~ {'a': 1994,  'b': 2006}   # ~50/50 — priors ignored
```

### Root cause
In `sklearn_genetic/space/space.py`, `Categorical.sample()` is:

```python
def sample(self):
    return self.rvs(self.choices)
```

`self.rvs` is either `numpy.random.Generator.choice` or `random.choice`, but the stored `self.priors` is never passed. For the numpy path it would need `self.rvs(self.choices, p=self.priors)`; `random.choice` does not support weights at all (`random.choices(..., weights=...)` would be needed).

### Impact
- The GA's initial population and any categorical mutation ignore user-supplied priors, so a documented feature is a no-op.
- No error is raised, so users have no signal that their priors are being dropped.

### Notes for whoever fixes this
- The two sampling backends differ: `self.rvs = self.rng.choice if self.rng else random.choice`. A fix needs to pass weights in both branches (`p=` for numpy, switch to `random.choices(..., weights=...)` for the stdlib branch), and handle the `distribution="choice"` case.
- Minor related observation in the same constructor: `self.rng = np.random.default_rng(random_state) if random_state else None` treats `random_state=0` as falsy, so seed `0` silently falls back to the stdlib RNG path rather than the numpy generator. Sampling stays deterministic (via `random.seed(0)`), but `0` behaves differently from other integer seeds.

I verified the above against current `master`. (Reporting only — not opening a PR for this one.)

delaidam:
Fixes #314

`Categorical.sample()` called `self.rvs(self.choices)` without ever passing
`self.priors` along, so priors were validated and stored on the object but
had zero effect on sampling, neaning that every choice was drawn with equal probability
regardless of documented weights. This affects both sampling backends
(`numpy.random.Generator.choice` when `random_state` is set, `random.choice`
otherwise), which need different call signatures to support weights
(`p=` vs. `random.choices(..., weights=...)`).

Also fixes a related bug in the same constructor: `random_state=0` was
treated as falsy (`None if not self.random_state else ...`), so seed `0`
silently fell back to the unseeded stdlib path instead of the numpy
generator.

## Changes
- `Categorical.sample()` now passes `self.priors` to the numpy `Generator.choice`
  path (`p=`), and uses `random.choices(..., weights=...)` for the stdlib path
  when priors are set.
- Fixed the `random_state=0` falsy check in `Categorical.__init__`.
- Added 3 regression tests: priors actually skew sampling on both backends
  (verified against a heavy 98/2 split, matching the issue's repro), and
  `random_state=0` correctly activates the numpy generator.

Ran `black --check` and the full suite locally: 361 passed, 4 skipped.

### DIFF
--- a/sklearn_genetic/space/space.py
+++ b/sklearn_genetic/space/space.py
@@ -184,7 +184,9 @@ class Categorical(BaseDimension):
         self.distribution = distribution
         self.random_state = random_state
         random.seed(random_state)
-        self.rng = None if not self.random_state else np.random.default_rng(self.random_state)
+        self.rng = (
+            np.random.default_rng(self.random_state) if self.random_state is not None else None
+        )
 
         if self.distribution == CategoricalDistributions.choice.value:
             self.rvs = self.rng.choice if self.rng else random.choice
@@ -198,7 +200,11 @@ class Categorical(BaseDimension):
     def sample(self) -> Any:
         """Sample a random value from the assigned distribution"""
 
-        return self.rvs(self.choices)
+        if self.priors is None:
+            return self.rvs(self.choices)
+        if self.rng is not None:
+            return self.rvs(self.choices, p=self.priors)
+        return random.choices(self.choices, weights=self.priors, k=1)[0]
 
 
 def check_space(param_grid: dict = None):
--- a/sklearn_genetic/space/tests/test_space.py
+++ b/sklearn_genetic/space/tests/test_space.py
@@ -1,3 +1,4 @@
+from collections import Counter
 from typing import Any
 
 import pytest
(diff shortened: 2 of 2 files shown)


rodrigo-arenas:
Thanks @delaidam, this looks good to me. I verified that this addresses #314: `Categorical.sample()` now applies `priors` on both sampling paths (`p=` for the NumPy generator and `weights=` via `random.choices` for the stdlib path), and the `random_state=0` edge case is fixed as well.

The added regression tests cover the important cases, and I verified the issue repro locally with the expected ~98/2 split. No requested changes from my side.

PR #326, opened 2026-07-06
Fix Integer and Continuous ignoring random_state=0

### LINKED ISSUE #322: Integer and Continuous ignore random_state=0
🟢 **Good first issue — quick fix**

`Integer` and `Continuous` ignore `random_state=0`. With seed `0` you get non-reproducible sampling, but every other seed works fine.

```python
a = [Integer(1, 1000, random_state=0).sample() for _ in range(5)]
b = [Integer(1, 1000, random_state=0).sample() for _ in range(5)]
print(a == b)  # False
```

Cause: in `space.py`, both use `None if not self.random_state else ...`, so `0` is treated as falsy and no RNG is created. `Categorical` already got this right in #320 with `is not None` — same fix on lines 58 and 119.

delaidam:
Fixes #322

Same falsy-zero bug already fixed for Categorical in #320:
`self.rng = None if not self.random_state else np.random.default_rng(self.random_state)`
treats seed `0` as falsy, so no numpy Generator is created and sampling
silently falls back to being unseeded and leads to reproducibility breaks specifically
for `random_state=0`.

## Changes
- Fixed the falsy-zero check in both `Integer.__init__` and `Continuous.__init__`
  (`is not None` instead of truthiness).
- Added parametrized regression tests covering both classes: `random_state=0`
  now activates the numpy generator, and produces a deterministic sample
  sequence across two independently constructed instances.

Ran `black --check` and the full suite locally: 368 passed, 4 skipped.

### DIFF
--- a/sklearn_genetic/space/space.py
+++ b/sklearn_genetic/space/space.py
@@ -55,7 +55,9 @@ class Integer(BaseDimension):
         self.upper = upper
         self.distribution = distribution
         self.random_state = random_state
-        self.rng = None if not self.random_state else np.random.default_rng(self.random_state)
+        self.rng = (
+            np.random.default_rng(self.random_state) if self.random_state is not None else None
+        )
 
         if self.distribution == IntegerDistributions.uniform.value:
             self.rvs = stats.randint.rvs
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, this looks good to me. The fix matches #322: `Integer` and `Continuous` now treat `random_state=0` as a real seed instead of falling back to the unseeded path.

I verified locally:
- `pytest sklearn_genetic/space/tests/test_space.py` -> 45 passed
- the issue repro is deterministic for both `Integer` and `Continuous`

CI is green as well. No requested changes from my side.

PR #328, opened 2026-07-07
Unify checkpoint serialization with a validated save/load contract (#297)

### LINKED ISSUE #297: [Intermediate] Unify checkpoint serialization between ModelCheckpoint and save/load

mayoka0:
Closes #297 — the follow-up you flagged when reopening after #313.

## What this does
Defines and validates the contract between the two serialization paths, in one place.

Before, they lived apart and were only related by convention:
- `save`/`load` use `GeneticEstimatorMixin._serializable_state()` (full snapshot, restored via `setattr`).
- `ModelCheckpoint` hand-built a separate constructor-compatible `estimator_state`.

Now:
- `_checkpoint_state()` moves onto `GeneticEstimatorMixin`, right next to `_serializable_state()` and `save`/`load`, so both serialization shapes live in `_base.py`.
- `ModelCheckpoint.on_step` just calls `estimator._checkpoint_state()` (the ~48 lines of key lists / builder leave the callback).

## The contract, validated
`test_checkpoint_state_honors_serialization_contract` runs for **both** `GASearchCV` and `GAFeatureSelectionCV` and asserts:
1. every checkpoint key is an `__init__` parameter (`⊆ get_params()`), and the checkpoint actually replays through `ClassName(**state)`;
2. the checkpoint is a subset of `_serializable_state()`.

So the two paths can't silently drift apart anymore — if someone adds a checkpoint key that isn't a constructor param, or that save/load doesn't carry, the test fails.

## Notes
- **Behavior unchanged** — same keys are persisted; this is a relocation + validation, not a format change. Existing checkpoint/resume and save/load tests still pass.
- The two `SimpleNamespace`-based callback tests now use real estimators, since state building lives on the estimator itself.
- I kept `estimator_state` an explicit, intentional subset rather than deriving it from all of `get_params()` — the full param set includes grouped-config objects (`evolution_config`, …) whose `__init__` re-resolves overlapping values, so a plain `Class(**all_params)` replay would be riskier. Happy to revisit if you'd prefer the derived approach.

Verified locally: `test_persistence_and_helpers.py` (14 passed) and the checkpoint/resume tests pass; `black --check` clean.

### DIFF
--- a/sklearn_genetic/_base.py
+++ b/sklearn_genetic/_base.py
@@ -19,12 +19,64 @@ def history_record(history, index):
 class GeneticEstimatorMixin:
     _volatile_pickle_attrs = {"toolbox", "_stats", "_pop", "_hof", "hof"}
 
+    # Contract for the two serialization paths (see #297):
+    #
+    # * ``_serializable_state()`` is the FULL snapshot used by ``save``/``load``.
+    #   It is restored with ``setattr`` onto an existing instance, so it may hold
+    #   non-constructor attributes (``X_``, ``cv_results_``, ``best_estimator_``…).
+    # * ``_checkpoint_state()`` is the CONSTRUCTOR-COMPATIBLE subset used by
+    #   ``ModelCheckpoint`` to resume a search. Every key is an ``__init__``
+    #   parameter, so it can be replayed as ``ClassName(**state)``, and it is a
+    #   strict subset of ``_serializable_state()``.
+    #
+    # The serialization contract test enforces that relationship (every
+    # checkpoint key is an __init__ parameter and lives in _serializable_state)
+    # so the two paths cannot silently drift apart.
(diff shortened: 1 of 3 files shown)


rodrigo-arenas:
Thanks, this looks good to me. This completes the follow-up for #297 cleanly: the checkpoint-specific state now lives on `GeneticEstimatorMixin` next to `_serializable_state()`, while preserving the constructor-compatible checkpoint contract and the `runtime_state["fitness_cache"]` behavior from #310.

CI is green and I do not see any requested changes.

PR #330, opened 2026-07-08
fix: preserve logbook and fit_stats_ across checkpoint resume

Hemanth-1354:
## Summary

Resuming a `GASearchCV`/`GAFeatureSelectionCV` fit from a `ModelCheckpoint`
silently dropped every pre-resume candidate from `cv_results_`/`history` and
reset `fit_stats_` counters to zero. Root cause was two-fold: `ModelCheckpoint`
was checkpointing the per-generation summary logbook instead of the
per-candidate logbook that `cv_results_`/`history` actually depend on, and
`_register()` unconditionally rebuilt `self.logbook` right after the (wrong)
restore anyway. `ModelCheckpoint` now persists the real per-candidate logbook
and `fit_stats_` in `runtime_state`, and `fit()` re-applies them after
`_register()` rebuilds the (unpicklable) toolbox/population/hof, so resumed
runs accumulate history and counters instead of losing them.

## Related Issue

Part of #299

Population/hall-of-fame continuation across resume, per-generation gen
numbering continuity, and RNG-stream continuation are not addressed here —
those need the design discussion the issue calls for. This PR fixes the
data-loss bug (logbook/fit_stats_ silently reset on resume); #299 should stay
open for the remaining scope.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] Other (describe):

## Checklist

- [x] Tests pass locally (`pytest sklearn_genetic/`) — 285 passed
- [x] Documentation updated if needed (CHANGELOG.md + docs-vitepress `latest` release notes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Full release notes with code examples are in the [documentation](https://sklearngeneticopt.rodrigo-arenas.com/).
 
+## Unreleased
+
+### Bug Fixes
+
+- Fixed `ModelCheckpoint` resume dropping prior-run history: `logbook`, `cv_results_`, and `history` were silently reset to only the post-resume generations, and `fit_stats_` counters were zeroed on every resume. `ModelCheckpoint` now persists the real per-candidate logbook and `fit_stats_`, and `GASearchCV`/`GAFeatureSelectionCV.fit` restore them after `_register()` rebuilds the (unpicklable) DEAP toolbox/population/hof (#299).
+
 ## 0.13.3
 
 ### New Features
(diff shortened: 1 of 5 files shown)


rodrigo-arenas:
Thanks for the update! I reviewed the implementation and this now covers the missing checkpoint resume state for `logbook`/`cv_results_`/`history` and `fit_stats_`. I also verified the focused resume tests locally, plus the analogous `GAFeatureSelectionCV` path.

We'll keep #299 open for the remaining full runtime/deterministic resume follow-up work.

PR #331, opened 2026-07-08
fix: continue generation numbering across checkpoint resume

Hemanth-1354:
## Summary

Every algorithm variant (`eaSimple`, `eaMuPlusLambda`, `eaMuCommaLambda`)
restarted its local `gen` counter at 0 on every `fit()` call, so a resumed
run's `logbook`/`history` showed duplicate generation indices (e.g. `0, 1`
from the first run followed by another `0, 1` after resume) instead of one
continuous, increasing sequence. Adds `_seed_logbook()`, which reuses the
checkpointed per-generation log and offsets new generation numbers past its
last recorded value, threaded through `GASearchCV`/`GAFeatureSelectionCV.fit`
via a new `resume_log` parameter on each algorithm function.

## Related Issue

Part of #299 (follow-up to the logbook/fit_stats_ fix in #330)

Population/hall-of-fame continuation across resume and RNG-stream
continuation are still not addressed, population/hof restart fresh each
resume (only the fitness cache and logbook carry over), and the RNG reseeds
from `random_state` rather than continuing the stream. Those need the design
discussion the issue calls for; #299 should stay open for that remaining
scope.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] Other (describe):

## Checklist

- [x] Tests pass locally (`pytest sklearn_genetic/`) — 286 passed
- [x] Documentation updated if needed

### DIFF
--- a/sklearn_genetic/_base.py
+++ b/sklearn_genetic/_base.py
@@ -122,6 +122,7 @@ class GeneticEstimatorMixin:
                 callbacks=self.callbacks,
                 verbose=self.verbose,
                 estimator=self,
+                resume_log=getattr(self, "_resume_generation_log", None),
             )
         else:
             raise ValueError(
--- a/sklearn_genetic/algorithms.py
+++ b/sklearn_genetic/algorithms.py
@@ -94,7 +94,7 @@ def _verbose_events(record):
     return ",".join(events) if events else "-"
 
 
-def _print_verbose_record(gen, nevals, record):
+def _print_verbose_record(gen, nevals, record, print_header=False):
     display_record = {
         "gen": gen,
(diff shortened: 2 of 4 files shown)


rodrigo-arenas:
Thanks! I reviewed this follow-up and the generation numbering now continues correctly when resuming from a checkpoint. CI is green, and I also verified the behavior locally across `GASearchCV`/`GAFeatureSelectionCV` and the three algorithm variants.

This looks good to merge. We'll keep #299 open for the remaining runtime-state resume scope around population/HOF continuation and RNG stream continuity.

PR #332, opened 2026-07-10
Preserve and correctly apply random_state across checkpoint resume

delaidam:
Part of #299

_checkpoint_core_keys did not include random_state, so a resumed
GASearchCV/GAFeatureSelectionCV silently fell back to whatever the
resuming instance's own constructor set (often None) instead of the
value the checkpointed run was built with.

Restoring the key alone was not sufficient either: _seed_global_rngs()
ran before the checkpoint's estimator_state was applied via
self.__dict__.update(), so it always seeded with the pre-resume
random_state regardless. Moved the seed call to after checkpoint
loading in both fit() methods.

Added a regression test that resumes twice from the same checkpoint
(with the resuming instances built without their own random_state) and
verifies identical results -- the only source of determinism available
to them is the seed restored from the checkpoint. Verified the test
fails without either half of the fix, and passes with both.

Ran black --check and the full suite locally: 374 passed, 4 skipped.

### DIFF
--- a/sklearn_genetic/_base.py
+++ b/sklearn_genetic/_base.py
@@ -41,6 +41,7 @@ class GeneticEstimatorMixin:
         "crossover_probability",
         "mutation_probability",
         "algorithm",
+        "random_state",
     )
     _checkpoint_optional_defaults = {
         "local_search": False,
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -1047,8 +1047,6 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             The callback is evaluated after fitting the estimators from the generation 1.
         """
 
-        _seed_global_rngs(self.random_state)
-
         self.X_ = X
         self.y_ = y
         self._n_iterations = self.generations + 1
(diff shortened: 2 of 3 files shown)


mayoka0 (approved):
Reviewed by pulling the branch and running the suite locally, not just reading the diff.

**The fix is correct on both fronts:**
1. Adding `random_state` to `_checkpoint_core_keys` — real gap, it was silently missing from the constructor-compatible subset.
2. Moving `_seed_global_rngs(self.random_state)` to *after* the checkpoint's `estimator_state` is applied via `self.__dict__.update(...)` — this is the subtler and more important half of the fix. Seeding before that point uses the pre-resume value even after (1) restores the correct one, so (1) alone wouldn't have actually fixed anything observable. Good catch.

**Verified locally:**
- `test_persistence_and_helpers.py` — 17 passed, including the new `test_checkpoint_resume_preserves_random_state`
- The new test's design is solid: two independent resumes from the same checkpoint (each built with `random_state=None` of their own) only match if the seed genuinely came from the checkpoint — that's a meaningful assertion, not just "it doesn't crash"
- `black --check` clean on all three touched files

Nice work — approving.

rodrigo-arenas:
Thanks! I reviewed this and the fix looks good: `random_state` is now preserved in the checkpoint state, and seeding happens after the checkpoint is restored, so resumed runs use the original run's seed.

This looks good to merge. We'll keep #299 open for the remaining full resume scope, especially true RNG stream continuation/population-HOF continuation.

PR #333, opened 2026-07-10
Fix Categorical.sample() returning numpy scalar types when seeded

### LINKED ISSUE #324: Seeded Categorical returns numpy scalars instead of Python types
🟠 **A bit harder — help wanted**

When `Categorical` has a `random_state`, `sample()` returns numpy types instead of plain Python ones. Without a seed it returns normal Python values, so the return type depends on whether a seed is set.

```python
from sklearn_genetic.space import Categorical
Categorical([True, False], random_state=1).sample()  # np.True_ (numpy bool)
Categorical([True, False]).sample()                  # True (python bool)
Categorical([1, 2, 3], random_state=1).sample()      # np.int64
```

These values end up in `best_params_` and get passed to the estimator, so checks like `x is True` or `isinstance(x, bool)` can behave unexpectedly. Not urgent, but a bit of a trap. Slightly more involved since it touches how values flow through, so tagging it as harder.

delaidam:
Fixes #324

A seeded Categorical sampled via self.rng.choice(self.choices, ...),
which numpy coerces to one of its own scalar types (np.True_, np.int64,
np.str_) instead of returning the original Python object -- unlike the
unseeded random-module path, which passed values through untouched. The
sampled value's type silently depended on whether random_state was set,
which is surprising for values that flow into best_params_ and get
passed to the estimator (e.g. 'x is True' checks).

sample() now always samples an index and looks it up in the original
choices list, so the exact stored Python object is returned regardless
of which RNG backend is used. This also let self.rvs be dropped, since
it's no longer needed by either backend.

Verified the new test fails on the previous implementation for every
seeded case (bool/int/str) and passes with the fix.

Ran black --check and the full suite locally: 379 passed, 4 skipped.

### DIFF
--- a/sklearn_genetic/space/space.py
+++ b/sklearn_genetic/space/space.py
@@ -192,9 +192,6 @@ class Categorical(BaseDimension):
             np.random.default_rng(self.random_state) if self.random_state is not None else None
         )
 
-        if self.distribution == CategoricalDistributions.choice.value:
-            self.rvs = self.rng.choice if self.rng else random.choice
-
     def __repr__(self) -> str:
         if self.priors is None:
             return f"{self.__class__.__name__}(choices={self.choices!r})"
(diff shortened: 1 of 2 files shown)


mayoka0 (approved):
Reviewed by pulling the branch, running the suite, and re-running the exact repro from the original issue.

**The fix:** sample an index via `rng.choice(n_choices, p=self.priors)` (numpy coerces the *index*, which gets normalized right back to a plain int via `int(index)`), then index into the original `self.choices` list to return the real Python object. That sidesteps the type-coercion problem entirely instead of trying to convert after the fact — clean approach.

**Verified locally:**
- `space/tests/test_space.py` — 51 passed, including the new parametrized `test_categorical_sample_returns_native_python_types` (covers bool/int/str, with/without priors, with/without seed — good matrix)
- Re-ran the original bug repro from #324 directly:
  ```
  Categorical([True, False], random_state=1).sample()  -> True   (type: bool)   ✓
  Categorical([1, 2, 3], random_state=1).sample()       -> 2      (type: int)    ✓
  Categorical(['a','b','c'], priors=[.5,.3,.2], random_state=1).sample() -> 'b'  (type: str) ✓
  ```
  All previously returned numpy scalar types; all now return the exact stored Python object.
- `black --check` clean

Approving.

rodrigo-arenas:
Thanks! I reviewed this and the approach looks good: sampling an index and then returning the original `choices` entry preserves the exact Python object/type while keeping seeded determinism and priors behavior intact.

This looks good to merge.

PR #334, opened 2026-07-11
Benchmark cache behavior under candidate collisions

### LINKED ISSUE #300: [Advanced] Benchmark and optimize duplicate candidate evaluation/cache behavior

ShiHuiwen-creat:
## Summary

- add an opt-in classification scenario with only four discrete candidates to make cache collisions reproducible
- report cache mode, cache hits, duplicate candidates, and actual cross-validation calls separately
- keep cache-on/cache-off comparisons compatible with older benchmark JSON files
- add focused tests for the collision search space and cache comparison reporting

## Verification

- `PYTHONPATH=. python -m pytest -q -o addopts=''`
- `python -m black --check benchmarks/benchmark_fit.py sklearn_genetic/tests/test_benchmark_fit.py`
- quick benchmark with cache enabled: 4 cross-validation calls, 21 cache hits
- quick benchmark with cache disabled: 26 cross-validation calls, 0 cache hits

Closes #300

### DIFF
--- a/benchmarks/benchmark_fit.py
+++ b/benchmarks/benchmark_fit.py
@@ -7,6 +7,10 @@ Run from the repository root, for example:
     python benchmarks/benchmark_fit.py --parallel-backends auto cv --runs 3
     python benchmarks/benchmark_fit.py --label current --output-json benchmarks/current.json
     python benchmarks/benchmark_fit.py --compare-json benchmarks/baseline.json
+    python benchmarks/benchmark_fit.py --quick --scenarios classification_lr_collisions \
+        --estimators gasearch --n-jobs 1 --label cache_on --use-cache
+    python benchmarks/benchmark_fit.py --quick --scenarios classification_lr_collisions \
+        --estimators gasearch --n-jobs 1 --label cache_off --no-use-cache
 
 The benchmark intentionally lives outside the test suite. It measures wall time,
 candidate evaluation counters, serial/parallel strategies, and holdout model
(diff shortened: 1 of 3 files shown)


rodrigo-arenas:
@ShiHuiwen-creat  Thanks for the benchmark work! The new collision scenario and cache counters look useful. I found one comparison issue before merge:

`comparison_key()` ignores `use_cache`, but `print_comparison_table()` builds `baseline_by_key` as a dict. If a baseline JSON has both cache-on and cache-off summaries for the same scenario/config, one row overwrites the other, so the comparison can pair `current_use_cache=True` with `baseline_use_cache=False`.

Could you adjust the matching logic so summaries with `use_cache` compare against the same cache mode, while still supporting old JSON files that do not have the `use_cache` field?

ShiHuiwen-creat:
Thanks for catching this — you're right that indexing only by the legacy comparison key could overwrite one cache mode.

I've updated the matching logic so baselines that include `use_cache` are indexed by both the existing configuration key and cache mode. Current rows now match the same cache mode exactly. For backward compatibility, baseline rows without `use_cache` are treated as historical cache-on results (the previous default), so only current cache-on rows can use that fallback; cache-off rows won't be compared across modes.

I added regression coverage for both cases:

- a baseline containing cache-on and cache-off rows matches each current row to the same mode
- an old baseline without `use_cache` remains usable for cache-on, but not cache-off

Local verification: 377 passed, 4 skipped; Black and `git diff --check` pass. The fix is in e6da40f.

rodrigo-arenas:
Thanks for the update! I re-reviewed the cache-mode comparison fix and it addresses the issue: cache-on and cache-off.

This looks good to merge.

PR #340, opened 2026-07-12
Add groups support to fit for group-aware cross-validation (#338)

### LINKED ISSUE #338: [Intermediate] Add groups support to fit for group-aware cross-validation
## Summary

Add explicit `groups` support to `GASearchCV.fit()` and `GAFeatureSelectionCV.fit()` so group-aware CV splitters can be used without precomputing splits manually.

## Why this helps

The public API currently mirrors `fit(X, y=None, callbacks=None)`, but scikit-learn users often pass `groups` when using splitters such as `GroupKFold`, `LeaveOneGroupOut`, or grouped custom CV objects. Today the fit loop materializes CV splits internally with `cv_orig.split(X, y)`, so there is no public way to pass `groups` through the estimator API.

Supporting `groups` would make the main estimators behave more like sklearn search estimators and avoid fragile workarounds where users precompute CV splits outside the estimator.

## Suggested implementation

Start by reading:

  - Review both `GASearchCV.fit` and `GAFeatureSelectionCV.fit`.
  - Find where `check_cv(...)` is called and where `_cv_splits` is materialized.

Possible approach:

- Extend the fit signatures to accept `groups=None`.
- Store `self.groups_` or another private fit-time attribute if needed.
- Pass `groups` to `cv_orig.split(X, y, groups)` when building `_cv_splits`.
- Ensure final-selection CV, if it reuses or creates CV splits, handles the same groups consistently.
- Keep `groups=None` behavior unchanged.

## Tests to add

Add focused tests for:

1. `GASearchCV.fit(X, y, groups=groups)` works with `GroupKFold` and never mixes groups across train/test folds
2. `GAFeatureSelectionCV.fit(X, y, groups=groups)` works with `GroupKFold`
3. existing behavior with `groups=None` remains unchanged
4. final selection, if enabled, uses group-aware splits or clearly documents/handles the limitation

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_genetic_search.py sklearn_genetic/tests/test_feature_selection.py
```

mayoka0:
Closes #338.

## What this does
`GASearchCV.fit` and `GAFeatureSelectionCV.fit` now accept `groups=None` and forward it to every place CV splits are materialized:

- the main `_cv_splits` construction in **both** estimators — both `cv_orig.split(...)` and `cv_orig.get_n_splits(...)` (the latter matters for splitters like `LeaveOneGroupOut`, whose `get_n_splits` requires groups);
- `_final_selection_splits`, so an explicit group-aware `final_selection_cv` receives the same groups as the main search.

The parameter is appended after `callbacks` so existing positional calls keep working, and `groups=None` passes through untouched — non-group splitters (`KFold`, `StratifiedKFold`, ints) ignore it, so default behavior is byte-for-byte unchanged.

## Why it matters
Before this, group-aware splitters could not be used at all: `GASearchCV(cv=GroupKFold(3), ...).fit(X, y)` raised `ValueError: The 'groups' parameter should not be None.` There was no public way to pass groups through, so users had to precompute splits manually.

## Tests
- `test_gasearch_fit_with_groups_supports_group_kfold` — fit succeeds under `GroupKFold` and **every materialized split keeps train/test groups disjoint** (asserted directly on `_cv_splits`, not just "no crash").
- `test_gasearch_final_selection_splits_use_groups` — full fit with `final_selection=True` and a `GroupKFold` `final_selection_cv`; final-selection splits are group-disjoint too.
- `test_feature_selection_fit_with_groups_supports_group_kfold` — same disjointness guarantee for `GAFeatureSelectionCV`, plus a valid support mask.
- `groups=None` unchanged: the full existing suites are the regression evidence (below).

## Verified locally
- New tests: 3 passed
- Full `test_genetic_search.py` + `test_feature_selection.py`: **107 passed**
- `test_persistence_and_helpers.py` (checkpoint/resume shares the fit flow): 17 passed
- `black --check` clean

### DIFF
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -933,7 +933,10 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             return self._cv_splits
 
         cv = check_cv(self.final_selection_cv, self.y_, classifier=_is_classifier(self.estimator))
-        return list(cv.split(self.X_, self.y_, groups=getattr(self, "groups_", None)))
+        groups = getattr(self, "groups_", None)
+        if groups is None:
+            return list(cv.split(self.X_, self.y_))
+        return list(cv.split(self.X_, self.y_, groups=groups))
 
     def _score_final_candidate(self, params, cv_splits):
         local_estimator = clone(self.estimator)
@@ -1148,6 +1151,10 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             self.metrics_list = self.scorer_.keys()
             self.multimetric_ = True
 
+        # Only pass groups when given, so custom splitters written without a
+        # groups argument keep working when no groups are used.
+        cv_groups_kwargs = {} if self.groups_ is None else {"groups": self.groups_}
+
         # Check cv and get the n_splits
         if _is_outlier_detector(self.estimator):
             # For outlier detectors, better to use KFold instead of classifier-based CV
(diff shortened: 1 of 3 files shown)


mayoka0:
Went over this again and found an edge case: a custom cv splitter written without a groups argument would break, even with no groups passed. Fixed — groups is now only forwarded when it's actually given, and there's a regression test for the old-style splitter.

rodrigo-arenas:
Thanks! I reviewed the groups support and the implementation looks good. 

The logic looks ready; the only remaining blocker I see is the failing `codecov/patch` check.

mayoka0:
Thanks! Added a test for the uncovered line — an explicit final_selection_cv with no groups. That was the only path nothing exercised. Coverage should be green now.

rodrigo-arenas:
Thanks for the update! I re-reviewed the latest version and the added coverage addresses the previous Codecov gap.

This looks good to merge.

PR #341, opened 2026-07-12
Forward sample_weight and fit params through candidate CV and refit (#339)

### LINKED ISSUE #339: [Intermediate] Forward sample_weight and fit params through candidate CV and refit
## Summary

Support passing estimator fit metadata such as `sample_weight` through `GASearchCV` and `GAFeatureSelectionCV` during candidate cross-validation and final refit.

## Why this helps

Many sklearn estimators accept `sample_weight` or other fit-time parameters. The current genetic search API evaluates candidates via `cross_validate(...)` and then refits the selected estimator with `estimator.fit(X, y)`, but the public `fit()` methods do not accept or forward fit parameters. This means users cannot run weighted searches through the main API even when the underlying estimator supports weights.

This is a common sklearn compatibility expectation for search estimators and matters for imbalanced datasets, cost-sensitive learning, and estimators with fit-time metadata.

## Suggested implementation

Start by reading:

  - Review `_evaluate_individual` for both `GASearchCV` and `GAFeatureSelectionCV`.
  - Review final refit blocks for both estimators.
  - Review population-level evaluation flow because candidate evaluation can run serially or in parallel.
- Relevant sklearn docs for `cross_validate` fit parameter handling in the supported sklearn versions.

Possible approach:

- Extend `fit()` to accept `**fit_params` or a targeted `sample_weight=None` parameter.
- Forward the metadata into `cross_validate(...)` during candidate evaluation.
- Forward the same metadata to the final `estimator.fit(...)` when `refit=True`.
- For `GAFeatureSelectionCV`, slice only `X`; keep metadata such as `sample_weight` aligned to samples, not features.
- Keep the default behavior unchanged when no metadata is passed.

## Tests to add

Add tests for:

1. `GASearchCV.fit(..., sample_weight=weights)` forwards weights into candidate CV fits
2. final refit receives `sample_weight` when `refit=True`
3. `GAFeatureSelectionCV` forwards `sample_weight` while still transforming only feature columns
4. unsupported fit params fail with a clear estimator/sklearn error rather than being silently ignored

A small custom estimator that records received `sample_weight` may make this easier and faster than relying on model scores.

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_genetic_search.py sklearn_genetic/tests/test_feature_selection.py
```

mayoka0:
Closes #339. **Builds on #340** (both extend the same `fit()` signatures — this branch includes that commit so the two don't conflict; merging #340 first will leave only this PR's own commit here).

## What this does
`GASearchCV.fit` and `GAFeatureSelectionCV.fit` now accept `**fit_params` (e.g. `sample_weight`) and forward them to **every** place the estimator is fitted:

1. **Candidate evaluation** — `cross_validate(..., params=self._fit_params)` in both `_evaluate_individual` implementations. scikit-learn slices sample-aligned params per CV fold (verified against sklearn 1.9; the `params=` kwarg exists throughout the supported `scikit-learn>=1.5` range).
2. **Final-selection re-scoring** — `_score_final_candidate` passes the same params, so a weighted search ranks its top-k candidates under the same weighting it searched with.
3. **Final refit** — both refit blocks pass `**fit_params` to `estimator.fit`.

For `GAFeatureSelectionCV`, only `X` is feature-sliced during the search; `sample_weight` stays aligned to samples and is row-sliced per fold by sklearn (tested explicitly).

**Parallel path:** params live on `self._fit_params`, and the population-parallel path in `evaluation.py` invokes `estimator._evaluate_individual` through joblib — the estimator (and its params) is what gets shipped to workers, so no separate threading was needed there.

## Behavior guarantees
- No fit params → `params=None` → byte-for-byte identical behavior (full suites pass).
- Unsupported params are **not** silently dropped: they surface as the estimator's own `TypeError` (tested with `error_score="raise"`).

## Tests
- `test_gasearch_forwards_sample_weight_to_candidate_cv_and_refit` — a recording estimator asserts every candidate fold fit got a **per-fold slice** (`< n_samples`) and the refit got the **full** vector (`== n_samples`).
- `test_gasearch_final_selection_scoring_receives_fit_params` — full run with `final_selection=True`; every fit in the pipeline was weighted.
- `test_feature_selection_forwards_sample_weight_and_slices_only_features` — same guarantees for `GAFeatureSelectionCV`, plus `X.shape[1]` never exceeds the original feature count.
- `test_gasearch_unsupported_fit_param_raises_clearly` — bogus param → `TypeError`.

## Verified locally
- New tests: 4 passed (plus the 3 groups tests from #340: 7 total)
- Full `test_genetic_search.py` + `test_feature_selection.py` + `test_persistence_and_helpers.py`: **128 passed**
- `black --check` clean

### DIFF
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -833,6 +833,7 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             pre_dispatch=self.pre_dispatch,
             error_score=self.error_score,
             return_train_score=self.return_train_score,
+            params=getattr(self, "_fit_params", None) or None,
         )
 
         cv_scores = cv_results[f"test_{self.refit_metric}"]
@@ -952,6 +953,7 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             pre_dispatch=self.pre_dispatch,
             error_score=self.error_score,
             return_train_score=False,
+            params=getattr(self, "_fit_params", None) or None,
         )
         cv_scores = cv_results[f"test_{self.refit_metric}"]
         return float(np.mean(cv_scores)), cv_scores
@@ -1030,7 +1032,7 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
 
         return selected_index, selected_score, selected_params
 
-    def fit(self, X, y=None, callbacks=None, groups=None):
+    def fit(self, X, y=None, callbacks=None, groups=None, **fit_params):
         """
         Main method of GASearchCV, starts the optimization
         procedure with the hyperparameters of the given estimator
@@ -1052,11 +1054,18 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             Group labels for the samples used while splitting the dataset into
             train/test sets. Only used in conjunction with a "Group" cv
             instance such as :class:`~sklearn.model_selection.GroupKFold`.
+        **fit_params : dict of str -> object
+            Parameters passed to the ``fit`` method of the underlying
+            estimator (e.g. ``sample_weight``). They are forwarded to every
+            candidate cross-validation, to final-selection scoring, and to
+            the final refit. Sample-aligned parameters are sliced per CV
+            fold by scikit-learn.
         """
 
         self.X_ = X
         self.y_ = y
         self.groups_ = groups
+        self._fit_params = fit_params
         self._n_iterations = self.generations + 1
         self.refit_metric = "score"
         self.multimetric_ = False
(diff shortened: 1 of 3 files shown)


mayoka0:
Double-checked the parallel path directly: same seed, weighted, serial vs population-parallel (n_jobs=2) give identical scores and fitness history, and weighted vs unweighted clearly differ. So weights really do reach the workers. Also pulled in the groups edge-case fix from #340.

rodrigo-arenas:
Thanks for the update! The implementation looks aligned with #339 and CI is green, but this branch now conflicts with `master` because #340 has already been merged while this PR still contains the #340 commits/merge commits.

Could you please rebase/update this branch on the latest `master` and resolve the conflicts in:

Ideally the PR should keep only the #339 fit-params/sample-weight changes after the rebase. Once updated, I’ll re-review it.

mayoka0:
Rebased onto the new master. Now it's just the sample_weight/fit_params changes, one commit, no leftover #340 content.

PR #346, opened 2026-07-13
fix(benchmark): make fit benchmark comparisons cache-mode aware

tusharui:
Problem

comparison_key in benchmarks/benchmark_fit.py did not include use_cache. When a baseline JSON contained both cache-on and cache-off rows for the same scenario/estimator/n_jobs/backend/initializer, one row silently overwrote the other in the lookup dict causing deltas to compare cache-on against cache-off.

Fix

- comparison_key now returns a 6-tuple including summary.get("use_cache")
- Removed redundant cache_aware_comparison_key (absorbed into comparison_key)
- print_comparison_table uses the new key for primary lookup, with a legacy 5-tuple fallback for old JSON files without use_cache

 All tests pass:
Screenshot (392)

Cache-on vs cache-off produces different comparison keys:
Screenshot 2026-07-14 012422

Comparison matches same cache mode (no cross-matching):
Screenshot 2026-07-14 012632

### DIFF
--- a/benchmarks/benchmark_fit.py
+++ b/benchmarks/benchmark_fit.py
@@ -744,36 +744,46 @@ def print_summary_table(summaries: list[dict[str, Any]]) -> None:
         print("\t".join(row))
 
 
-def comparison_key(summary: dict[str, Any]) -> tuple[str, str, str, str, str]:
+def comparison_key(summary: dict[str, Any]) -> tuple[str, str, str, str, str, bool | None]:
     return (
         summary["scenario"],
         summary["estimator"],
         str(summary["n_jobs"]),
         summary["parallel_backend"],
         summary["population_initializer"],
+        summary.get("use_cache"),
     )
 
 
-def cache_aware_comparison_key(
-    summary: dict[str, Any],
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, this is very close. I found one fallback edge case before merge: legacy baselines without `use_cache` are reported as `baseline_use_cache=True`, but the new fallback currently also matches them against `current_use_cache=False` rows.

That means a current cache-off run can still be compared against an old cache-on-style baseline. Could you keep the legacy fallback limited to `summary.get("use_cache", True) is True` (or otherwise treat missing baseline `use_cache` as cache-on only), and add a regression test where current has both cache modes but the baseline lacks `use_cache`, expecting only the cache-on row to compare?

The focused benchmark tests pass otherwise.

tusharui:
All 7 tests pass:
Addressed the fallback edge case in commit 37cc8ac.

Legacy baselines without use_cache are now only matched against
current_use_cache=True rows. Cache-off current rows are excluded.
Screenshot 2026-07-14 014746

rodrigo-arenas:
Thanks, it looks ready to merge

PR #347, opened 2026-07-13
fix: _select_final_candidate ignores criteria="min" -selects worst candidate for minimization

tusharui:
Summary

_select_final_candidate in GASearchCV always uses max() to pick the best re-evaluated candidate, even when criteria="min". For minimization problems (e.g., tuning toward RMSE or MAE), this silently selects the worst candidate.

Reproduction

from sklearn.tree import DecisionTreeClassifier
from sklearn_genetic import GASearchCV

estimator = GASearchCV(
    DecisionTreeClassifier(),
    param_grid={"max_depth": Integer(1, 5)},
    final_selection=True,
    criteria="min",
)
# After fit(), the final candidate is the one with the HIGHEST score
# instead of the lowest — opposite of what criteria="min" means.

Root cause

genetic_search.py:1010:
selected = max(candidate_results, key=lambda item: item["score"])

Should be conditional on self.criteria.

Fix
if self.criteria == "min":
    selected = min(candidate_results, key=lambda item: item["score"])
else:
    selected = max(candidate_results, key=lambda item: item["score"])

TEST CASE 
Screenshot 2026-07-14 022630

### DIFF
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -1007,7 +1007,10 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
             )
 
         if candidate_results:
-            selected = max(candidate_results, key=lambda item: item["score"])
+            if self.criteria == "min":
+                selected = min(candidate_results, key=lambda item: item["score"])
+            else:
+                selected = max(candidate_results, key=lambda item: item["score"])
             selected_index = selected["index"]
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks! This fix looks good. `_select_final_candidate` now respects `criteria="min"` when choosing among re-evaluated final-selection candidates, and the regression test covers the minimization path directly.

Looks ready to merge.

PR #349, opened 2026-07-14
Detect duplicate benchmark comparison keys instead of silently overwriting

### LINKED ISSUE #336: [Intermediate] Detect duplicate benchmark comparison keys
## Summary

Make benchmark comparison helpers detect duplicate comparison keys instead of silently overwriting rows.

## Why this helps

Both benchmark scripts build a lookup with a dictionary keyed by a small subset of summary fields. If two summaries share the same comparison key, the later row silently replaces the earlier one. That can hide benchmark rows and produce misleading comparisons, especially as new dimensions are added to benchmark summaries.

This is most visible in comparison helpers such as:

- `benchmarks/benchmark_fit.py::print_comparison_table`
- `benchmarks/benchmark_search_methods.py::print_comparison_table`

## Suggested implementation

Start by reading:

- `benchmarks/benchmark_fit.py`
  - Review `comparison_key` and `print_comparison_table`.
- `benchmarks/benchmark_search_methods.py`
  - Review the equivalent comparison logic.

Add a small helper that builds the baseline lookup and either:

- raises a clear `ValueError` when duplicate comparison keys are present; or
- prints a clear warning and skips ambiguous matches.

Prefer a behavior that prevents silently wrong deltas. Keep the output backward-compatible for non-duplicate benchmark JSON files.

## Tests to add

Add focused tests with synthetic summary rows for:

1. duplicate baseline keys are detected and do not overwrite silently
2. non-duplicate baselines still compare as before
3. both `benchmark_fit.py` and `benchmark_search_methods.py` use the same duplicate-key behavior or a shared helper

## Validation

Run:

```bash
pytest sklearn_genetic/tests/test_benchmark_fit.py
pytest sklearn_genetic/tests/test_benchmark_search_methods.py
python benchmarks/benchmark_search_methods.py --quick
```

If `test_benchmark_search_methods.py` does not exist yet, add a small focused test module for the helper logic.

delaidam:
Closes #336

Both benchmark_fit.py::print_comparison_table and
benchmark_search_methods.py::print_comparison_table built their baseline
lookup as {comparison_key(summary): summary for summary in baseline}, so
two baseline rows sharing a comparison key silently picked whichever was
built last -- comparing the current run against the wrong baseline row
with no indication in the printed table.

Added a shared build_baseline_lookup() helper in benchmark_fit.py
(benchmark_search_methods.py already imports shared code from there) that
raises a clear ValueError on a duplicate key. Both scripts'
print_comparison_table now use it, including the legacy (no use_cache)
baseline lookup in benchmark_fit.py.

Added sklearn_genetic/tests/test_benchmark_search_methods.py, which did
not exist before -- it needs a small sys.path shim since
benchmark_search_methods.py's own sibling-relative 'from benchmark_fit
import ...' only resolves when benchmarks/ itself is on sys.path (true
when run directly, not when imported as benchmarks.* for testing).

Verified all four new duplicate-key tests fail against the previous
silent-overwrite implementation and pass with the fix. Ran black --check,
the full suite (405 passed, 4 skipped), and a real
benchmark_search_methods.py --quick run to confirm the script still works.

### DIFF
--- a/benchmarks/benchmark_fit.py
+++ b/benchmarks/benchmark_fit.py
@@ -755,35 +755,50 @@ def comparison_key(summary: dict[str, Any]) -> tuple[str, str, str, str, str, bo
     )
 
 
-def print_comparison_table(current: list[dict[str, Any]], baseline: list[dict[str, Any]]) -> None:
-    baseline_by_key: dict[tuple, dict[str, Any]] = {}
-    for summary in baseline:
-        baseline_by_key[comparison_key(summary)] = summary
+def _legacy_comparison_key(summary: dict[str, Any]) -> tuple[str, str, str, str, str]:
+    return (
+        summary["scenario"],
+        summary["estimator"],
+        str(summary["n_jobs"]),
+        summary["parallel_backend"],
+        summary["population_initializer"],
+    )
 
-    legacy_baseline_by_key: dict[tuple, dict[str, Any]] = {}
+
+def build_baseline_lookup(
+    baseline: list[dict[str, Any]], *, key_func: Callable[[dict[str, Any]], tuple]
+) -> dict[tuple, dict[str, Any]]:
+    """Build a ``{key_func(summary): summary}`` lookup.
+
+    Raises rather than silently letting a later row overwrite an earlier one
+    with the same key -- a silent overwrite would compare against the wrong
+    baseline row with no indication in the printed table (#336).
+    """
+    lookup: dict[tuple, dict[str, Any]] = {}
     for summary in baseline:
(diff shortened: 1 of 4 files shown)


rodrigo-arenas:
Thanks for the contribution!

I reviewed the PR and it looks complete to me. The shared `build_baseline_lookup()` helper addresses the silent overwrite problem described in #336, and both benchmark comparison paths now fail clearly when duplicate baseline keys are present.

Everything passed, and the GitHub checks are green as well. I don't have any requested changes.

PR #350, opened 2026-07-17
fix: replace broad exception swallowing with specific exceptions and logging

tusharui:
## Problem

`save()`, `load()`, and `ModelCheckpoint` used bare `except Exception` with `print()`, silently swallowing all errors (including `PermissionError`, `MemoryError`, `OSError`). Users had no way to know if save/load failed.

## Changes

- **`_base.py`**: `save()`/`load()` now catch specific exceptions (`OSError`, `pickle.PicklingError`, `FileNotFoundError`, `KeyError`) and re-raise after logging
- **`model_checkpoint.py`**: `on_step()` catches only `OSError`/`pickle.PicklingError` (checkpoint failure shouldn't crash optimization); `load()` re-raises after logging
- All `print()` calls replaced with Python `logging` module (`logger.info`/`error`/`warning`)
- Tests updated to use `caplog` instead of `capsys` for log assertions
Screenshot 2026-07-17 153220

### DIFF
--- a/sklearn_genetic/_base.py
+++ b/sklearn_genetic/_base.py
@@ -1,3 +1,4 @@
+import logging
 import pickle
 
 from sklearn.exceptions import NotFittedError
@@ -5,6 +6,8 @@ from sklearn.utils.validation import check_is_fitted
 
 from .algorithms import algorithms_factory
 from .parameters import Algorithms
+
+logger = logging.getLogger(__name__)
 
 
 def reset_adapters(estimator):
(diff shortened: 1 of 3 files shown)


rodrigo-arenas:
Thanks for the PR! The direction looks good overall, and I verified the focused persistence tests locally. The remote checks are green too.

I found one behavior gap that I think we should fix before merging. `ModelCheckpoint.on_step()` currently catches only `OSError` and `pickle.PicklingError` around `pickle.dump()`, but pickle can also raise `AttributeError` or `TypeError` for non-serializable state. In that case the checkpoint callback still crashes the optimization, even though the PR description says checkpoint failures should not crash the run.

I reproduced this with an estimator whose `_checkpoint_state()` contains a lambda; `ModelCheckpoint.on_step()` raises `AttributeError` instead of logging a warning and returning. Could you please handle the same serialization exception family here that `save()` handles, and add a regression test for that path?

Minor test cleanup: `test_persistence_and_helpers.py` currently defines `test_checkpoint_load_raises_on_oserror` twice, so the later definition overwrites the earlier one. Could you rename one of them so both cases are collected explicitly?

After those two updates, this should be in good shape.

rodrigo-arenas:
Thanks for the updates, they look good! Two more things before merging:

1. The **Lint** check is red — `black --check .` wants to reformat `test_persistence_and_helpers.py` (missing a blank line before `test_checkpoint_resume_restores_adapter_state`). Running `black .` should fix it.
2. In `ModelCheckpoint.on_step()` (model_checkpoint.py), merging in `master` left the `estimator_state`/`runtime_state`/`checkpoint_data` computation duplicated: once outside the `try` (without `adapter_state`, and never used) and again inside the `try` (the one that actually counts). Could you remove the duplicated block outside the `try`?

Should be good to merge after that.

rodrigo-arenas:
Looks complete and ready to merge. Thank you!

rodrigo-arenas (approved):
Looks complete and ready to merge. Thank you!

PR #353, opened 2026-07-20
fix: restore adapter state on checkpoint resume

tusharui:
## Problem

When using `ModelCheckpoint` to resume a search, the crossover/mutation adapter step counters (`current_step`, `current_value`) were not saved or restored. This caused adaptive probability schedules (`ExponentialAdapter`, `InverseAdapter`, `PotentialAdapter`) to restart from step 0 instead of continuing from where the previous run left off.

## Reproducible Example

from sklearn.datasets import load_iris
from sklearn.tree import DecisionTreeClassifier
from sklearn_genetic import GASearchCV
from sklearn_genetic.callbacks.model_checkpoint import ModelCheckpoint
from sklearn_genetic.schedules.schedulers import ExponentialAdapter

X, y = load_iris(return_X_y=True)

def build():
    return GASearchCV(
        estimator=DecisionTreeClassifier(random_state=0),
        param_grid={"max_depth": Integer(1, 3)},
        cv=2, scoring="accuracy", population_size=4, generations=3,
        crossover_probability=ExponentialAdapter(0.8, 0.2, 0.1),
        mutation_probability=ExponentialAdapter(0.6, 0.1, 0.1),
    )

first = build()
first.fit(X, y, callbacks=ModelCheckpoint(checkpoint_path="cp.pkl"))
print(first.crossover_adapter.current_step)  # 3

resumed = build()
resumed.fit(X, y, callbacks=ModelCheckpoint(checkpoint_path="cp.pkl"))
print(resumed.crossover_adapter.current_step)
# Expected: 6  (3 + 3 more generations)
# Before fix: 3  (restarted from 0)
## Root Cause
ModelCheckpoint.on_step() saves _checkpoint_state() (constructor params) and runtime_state (fitness cache, logbook, fit stats) — but not the adapter objects. On resume, __dict__.update() restores constructor params while _reset_adapters() is skipped (checkpoint_loaded=True), leaving adapters stuck at current_step=0.
Fix
- schedules/base.py — Add state_dict() / load_state_dict() to BaseAdapter for serializing mutable state.
- callbacks/model_checkpoint.py — Save adapter state into runtime_state["adapter_state"].
- genetic_search.py — Restore adapter state on resume in both GASearchCV.fit() and GAFeatureSelectionCV.fit().
- Tests — 3 new tests covering adapter checkpoint resume for both estimators and state_dict round-trip.
Backward Compatibility
Old checkpoints without adapter_state in runtime_state still work — adapters default to step 0 (same behavior as before this fix).
Screenshot (400)

### DIFF
--- a/sklearn_genetic/callbacks/model_checkpoint.py
+++ b/sklearn_genetic/callbacks/model_checkpoint.py
@@ -38,6 +38,13 @@ class ModelCheckpoint(BaseCallback):
                 "fit_stats_": deepcopy(getattr(estimator, "fit_stats_", None)),
                 "candidate_logbook": deepcopy(getattr(estimator, "logbook", None)),
             }
+            crossover_adapter = getattr(estimator, "crossover_adapter", None)
+            mutation_adapter = getattr(estimator, "mutation_adapter", None)
+            if crossover_adapter is not None and mutation_adapter is not None:
+                runtime_state["adapter_state"] = {
+                    "crossover": crossover_adapter.state_dict(),
+                    "mutation": mutation_adapter.state_dict(),
+                }
             checkpoint_data = {
                 "estimator_state": estimator_state,
                 "logbook": deepcopy(logbook),
--- a/sklearn_genetic/genetic_search.py
+++ b/sklearn_genetic/genetic_search.py
@@ -1085,6 +1085,7 @@ class GASearchCV(GeneticEstimatorMixin, BaseSearchCV):
         restored_logbook = None
         restored_fit_stats = None
         restored_generation_log = None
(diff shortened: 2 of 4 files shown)


rodrigo-arenas:
Thanks for the contribution and the clear write-up, @tusharui! Reviewed the fix and it correctly addresses the root cause — adapters are restored via `load_state_dict()` onto the existing instances after resume, with sensible backward compatibility for old checkpoints without `adapter_state`. Verified locally: all tests pass, including the new adapter-resume coverage for both `GASearchCV` and `GAFeatureSelectionCV`. Looks good to merge.

PR #354, opened 2026-07-20
fix: remove global RNG seed side effect from Categorical.__init__

tusharui:
## Problem

Categorical.__init__ called  random.seed(random_state) which unconditionally mutated the global Python random module state. Integer and Continuous do not have this side effect.

## Fix

Removed the  random.seed(random_state)  line. When random_state is provided, `self.rng` (numpy `Generator`) is already used for sampling. When `None`, the global RNG is used as-is.

## Before/After

python
# Before: constructing this mutated global random state
Categorical(["a", "b"], random_state=42)  # calls random.seed(42)

# After: no side effect, sampling still deterministic via self.rng
Categorical(["a", "b"], random_state=42)  # uses np.random.default_rng(42)

### DIFF
--- a/sklearn_genetic/space/space.py
+++ b/sklearn_genetic/space/space.py
@@ -187,7 +187,6 @@ class Categorical(BaseDimension):
         self.choices = choices
         self.distribution = distribution
         self.random_state = random_state
-        random.seed(random_state)
         self.rng = (
             np.random.default_rng(self.random_state) if self.random_state is not None else None
         )


rodrigo-arenas:
Thanks for catching this and for the clean, minimal fix, @tusharui! Confirmed the root cause: this constructor-time `random.seed()` call was pure global-state pollution — when `random_state` is set, `sample()` already uses `self.rng` (numpy) instead of the stdlib `random` module, and when it's `None`, `random.seed(None)` never provided determinism anyway. It's also consistent with the rest of the library's reproducibility design: `_seed_global_rngs()` in `genetic_search.py` is the single point that seeds the global RNGs at `fit()` time, so individual space dimensions shouldn't be mutating global state on their own — this brings `Categorical` in line with `Integer`/`Continuous`, which never did this. Ran the full `space` test suite locally against this branch: all 51 tests pass, including the existing seeded/unseeded sampling coverage. Looks good to merge.

Issue #267, opened 2026-06-28
[DOCS] Share community-written articles and tutorials


rodrigo-arenas:
## Summary

We would love to collect community-written tutorials, blog posts, notebooks, and articles about using `sklearn-genetic-opt`.

Articles can live outside the project docs, for example on Medium, Dev.to, a personal blog, LinkedIn, GitHub Pages, or another public writing platform. After publishing, contributors can open a small pull request adding the link to the Community Articles page.

## Why this helps

External articles help new users learn from real workflows and give contributors public recognition for teaching others. The project docs stay focused on official documentation, while this page makes community writing easier to discover.

## Suggested article topics

- Hyperparameter tuning with `GASearchCV`
- Feature selection with `GAFeatureSelectionCV`
- Using `sklearn-genetic-opt` with sklearn Pipelines
- Comparing genetic search with `GridSearchCV` or `RandomizedSearchCV`
- Practical tips for designing search spaces
- Real-world examples with imbalanced data, regression, or model selection

## Suggested implementation

Start by reading this file:

- `docs-vitepress/versions/latest/community/articles.md`
  - This page lists external community articles in submission order.

To add your article:

1. Publish the article externally.
2. Add a new entry at the bottom of the `## Articles` list.
3. Include the article title, author name or GitHub handle, publication platform, link, and a short description.
4. Keep the entry concise so the page stays easy to scan.

Please append new entries below the existing ones. The list is ordered by submission order, not by ranking or preference.

Links should be relevant to `sklearn-genetic-opt`, respectful, and not primarily promotional.

## Tests to add

This is a documentation-only contribution, so no Python tests are required.

## Validation

Run:

```bash
node node_modules/vitepress/bin/vitepress.js build .
```

from the `docs-vitepress` directory.

## Notes for contributors

This is a good contribution for people who enjoy writing, teaching, or sharing applied machine learning workflows. The full article stays on your own platform; the repository only needs a small PR adding the link.

Akash-reddy-13:
Hi! I would like to contribute to this issue.

I am interested in writing a tutorial comparing traditional hyperparameter tuning methods with genetic optimization using sklearn-genetic-opt. Please let me know if this topic would be suitable.

Thanks!

rodrigo-arenas:
> Hi! I would like to contribute to this issue.
> 
[...]

PR #355, opened 2026-07-21
Add benchmark write-up: GridSearchCV vs GASearchCV comparison

Ayushaman1291:
## Summary
Adds a benchmark write-up comparing `GridSearchCV`, `RandomizedSearchCV`, and
`GASearchCV`, using the repository's own `benchmark_search_methods.py` script
rather than a custom comparison. Covers two scenarios (classification_lr,
regression_ridge) at two grid densities, with an honest breakdown of when
genetic search's advantage does and doesn't show up based on the actual
numbers.

## Related Issue
Addresses #267

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Refactor / cleanup
- [ ] Other (describe):

## Checklist
- [x] Tests pass locally (N/A — markdown-only change, no code modified)
- [x] Code is formatted with Black (N/A — markdown-only change)
- [ ] New functionality is covered by tests
- [x] Documentation updated if needed

### DIFF
--- a/benchmarks/gridsearch_vs_gasearch_writeup.md
+++ b/benchmarks/gridsearch_vs_gasearch_writeup.md
@@ -1,0 +1,128 @@
+# GridSearchCV vs GASearchCV: When Does Genetic Search Actually Pay Off?
+
+Hyperparameter tuning usually comes down to a tradeoff between how thoroughly you search
+and how much time you're willing to spend. `GridSearchCV` searches exhaustively; genetic
+algorithm-based approaches like `GASearchCV` (from
+[sklearn-genetic-opt](https://github.com/rodrigo-arenas/Sklearn-genetic-opt)) search
+adaptively, using evolutionary principles to focus evaluations on promising regions of
+the hyperparameter space.
+
+The common pitch for genetic search is "fewer evaluations, comparable results." I wanted
+to check that claim directly, using the project's own benchmarking tool
+(`benchmarks/benchmark_search_methods.py`) rather than a hand-rolled comparison — so the
+numbers below come straight from that script, unmodified.
+
+## Setup
+
+Two scenarios, both built into the benchmark suite:
+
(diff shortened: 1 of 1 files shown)


rodrigo-arenas:
Thanks for putting this together and for using the repo's own benchmark script instead of a custom one — the numbers reproduce locally and the methodology (matched CV, matched random_state, RandomizedSearchCV's n_iter aligned to GA's candidate budget) is sound.

That said, I don't think the conclusion is representative yet:

1. **Scenario selection stacks the deck against GA.** Of the 8 scenarios already defined in `benchmark_fit.py`, `classification_lr` and `regression_ridge` are the only two with just 2 hyperparameters (one continuous, one binary categorical) — the simplest, lowest-dimensional cases in the suite, and the ones structurally most favorable to exhaustive grid search. The suite also has `classification_rf` (6 params), `classification_gbm` (7 params), and their synthetic variants — exactly the high-dimensional, interaction-heavy cases the write-up's own "honest takeaway" section argues GA should win on. With `--grid-points 5`, `classification_rf` alone would need ~9,375 grid combinations vs GA's ~252 candidates. None of those scenarios are run, so the "GA doesn't win" conclusion is drawn only from the two cases where it structurally can't. Could you add at least one higher-dimensional scenario (e.g. `classification_rf` or `classification_gbm`) so the write-up backs up its own claim with data instead of prose?

2. **This doesn't match what issue #267 asks for.** #267 is about linking externally-published community articles on `docs-vitepress/versions/latest/community/articles.md`, not adding a new benchmark file under `benchmarks/`. Could you clarify how this PR maps to that issue, or should it target something else?

Happy to take another look once those are addressed.

rodrigo-arenas:
This still needs the previously requested changes: align the PR with issue #267, or retarget it, and add at least one higher-dimensional benchmark scenario.

PR #356, opened 2026-07-22
fix: improve ModelCheckpoint error handling and logging

tusharui:
## Summary

- Replace `print()` with `logging.getLogger(__name__)` for structured output
- Replace bare `except Exception` with specific exception types (`OSError`, `pickle.PicklingError`, `AttributeError`, `TypeError`)
- Separate `logbook_saver` call from pickle write `try/except` (separate concerns)
- Re-raise exceptions in `load()` instead of silently returning `None`

## Why This Change

`ModelCheckpoint` had several code quality issues:
1. Used `print()` instead of `logging` — inconsistent with the rest of the codebase
2. Bare `except Exception` swallowed all errors including `TypeError` which should propagate
3. `load()` silently returned `None` on any error, making debugging difficult — callers had no way to know if a checkpoint was corrupted vs missing vs permission denied

## Changes

**`sklearn_genetic/callbacks/model_checkpoint.py`**: Core fix — logging, specific exceptions, load() re-raise

**`sklearn_genetic/tests/test_persistence_and_helpers.py`**: Updated 2 tests to use `caplog` instead of `capsys`, and assert `load()` raises `FileNotFoundError` on missing files instead of returning `None`

### DIFF
--- a/sklearn_genetic/callbacks/model_checkpoint.py
+++ b/sklearn_genetic/callbacks/model_checkpoint.py
@@ -1,7 +1,11 @@
+import logging
 import pickle
+
 from .base import BaseCallback
 from .loggers import LogbookSaver
 from copy import deepcopy
+
+logger = logging.getLogger(__name__)
 
 
 class ModelCheckpoint(BaseCallback):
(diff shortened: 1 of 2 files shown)


rodrigo-arenas:
Thanks, this is a clean, well-scoped fix — no leftover duplication here.

The `codecov/patch` check is failing (70% patch coverage, 6 lines missing). The gap is in `load()`: the two new except blocks are never exercised by a test:

```python
except (pickle.UnpicklingError, EOFError, ValueError) as e:   # corrupted file
    ...
except OSError as e:                                           # e.g. PermissionError
    ...
```

The tests you updated only cover `FileNotFoundError` (missing checkpoint) and the `OSError` path in `on_step()` (missing directory on save). Could you add a couple of tests for `load()`:
- one that writes garbage bytes to the checkpoint file and asserts `load()` raises `pickle.UnpicklingError` (or similar) and logs "corrupted file"
- one that patches `builtins.open` to raise `PermissionError` and asserts `load()` re-raises it and logs the error

That should close the coverage gap and get the check green.

rodrigo-arenas:
This is now covered by #350, which has been merged. Please close this PR, or rebase and keep only any remaining distinct change.

PR #358, opened 2026-07-22
fix: add __getstate__/__setstate__ for proper pickling

tusharui:
## Summary

- Add `__getstate__` to `GeneticEstimatorMixin` that excludes unpicklable DEAP internals (`toolbox`, `_stats`, `_pop`, `_hof`) from the serialized state
- Add `__setstate__` to restore instance state cleanly, leaving DEAP attrs unset so they're rebuilt on `fit()`
- Remove `hof` from `_volatile_pickle_attrs` — it's a regular dict that should be serializable
- Update `save()`/`load()` to use `__getstate__`/`__setstate__`
- `load()` handles both the legacy dict format (with `estimator_state` key) and raw pickled instances for backward compatibility

## Why This Change

The old `save()`/`load()` used raw pickle on `__dict__` with manual filtering via `_serializable_state()`. Without proper `__getstate__`/`__setstate__`, `pickle.dumps(self)` and `joblib.dump()` would try to serialize unpicklable DEAP internals (`toolbox`, `_stats`, `_pop`, `_hof`), leading to errors or silent corruption. A proper pickle protocol makes the serialization contract explicit and prevents accidental serialization of DEAP objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Full release notes with code examples are in the [documentation](https://sklearn
 
 ## Unreleased
 
+### Bug Fixes
+
+- Fixed pickling of fitted `GASearchCV` and `GAFeatureSelectionCV` instances. `pickle.dumps`/`pickle.loads` and `joblib.dump`/`joblib.load` previously failed on the DEAP toolbox captured in `__dict__` (`Can't pickle <function ...>`) and now exclude the unpicklable DEAP internals (`toolbox`, `_stats`, `_pop`, `_hof`) via `__getstate__`/`__setstate__`. `save()`/`load()` use the same protocol, and `load()` also accepts files written directly with `pickle.dump(self, f)` (#358).
 ## 0.13.4
 
 ### New Features
--- a/sklearn_genetic/_base.py
+++ b/sklearn_genetic/_base.py
@@ -17,7 +17,7 @@ def history_record(history, index):
 
 
 class GeneticEstimatorMixin:
-    _volatile_pickle_attrs = {"toolbox", "_stats", "_pop", "_hof", "hof"}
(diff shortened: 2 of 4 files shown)


rodrigo-arenas:
This is a real fix for a real bug, not tied to an issue but worth having: on current master, `pickle.dumps()` and `joblib.dump()` both fail on a fitted `GASearchCV` (`Can't pickle <function GASearchCV._register.<locals>.<lambda>>`), since the DEAP toolbox ends up in `__dict__`. That matters beyond the project's own `.save()`/`.load()` API, since `joblib.dump` is the standard way sklearn users persist estimators. I verified locally that with this PR, `pickle.dumps/loads`, `joblib.dump/load`, refit after restore, and the legacy `save()`/`load()` path all work correctly for both `GASearchCV` and `GAFeatureSelectionCV`.

Two things needed before merging:

1. **No tests were added.** `__getstate__`/`__setstate__` and the new backward-compatible branch in `load()` (dict format vs. raw pickled instance) aren't covered by any test — that's also why `codecov/patch` is failing (the `else: self.__setstate__(checkpoint_data)` branch is the one missing line). Could you add tests for:
   - `pickle.dumps`/`pickle.loads` round-trip on a fitted `GASearchCV` and `GAFeatureSelectionCV`, asserting `predict`/`best_features_` still work afterward
   - a `joblib.dump`/`joblib.load` round-trip
   - the raw-pickled-instance branch in `load()` (i.e. loading a file produced by `pickle.dump(self, f)` directly, not through `save()`)
   - refit after restore, to confirm the toolbox rebuilds cleanly

2. **CHANGELOG.md** — `## Unreleased` is currently empty; this is exactly the kind of bug fix that belongs there.

rodrigo-arenas:
Please rebase on current master and resolve the conflicts with #350, keeping the new logging/re-raise behavior in save/load.

PR #370, opened 2026-07-30
Fail clearly on unsupported fit metadata (#364)

### LINKED ISSUE #364: [Advanced] Support sklearn metadata routing for groups and sample_weight
## Summary

Modern sklearn estimators increasingly rely on metadata such as `groups` and `sample_weight`. `GASearchCV` and `GAFeatureSelectionCV` should have a clearer compatibility story for forwarding metadata through cross-validation, scoring, and final refit.

## Suggested scope

- Audit how `fit` currently handles metadata-like inputs for both estimators.
- Add support for forwarding `groups` to CV splitting/scoring where appropriate.
- Add support for `sample_weight` during estimator fitting when the wrapped estimator accepts it.
- Align behavior with sklearn conventions as much as practical for the supported sklearn versions.
- Add tests with a grouped CV splitter and a sample-weight-aware estimator.
- Document supported metadata and any limitations.

## Acceptance criteria

- Group-aware CV can be used without silently dropping `groups`.
- `sample_weight` reaches compatible wrapped estimators during candidate evaluation and final refit.
- Unsupported metadata fails with a clear message instead of being ignored.
- Existing public APIs remain backward compatible.

## Pointers

Difficulty: Advanced

mayoka0:
Closes #364.

Most of #364 already landed in #340/#341: `groups` reaches group-aware CV splitters, and `sample_weight` (and other fit metadata) reaches candidate evaluation, final-selection scoring, and the final refit. The one acceptance criterion still open was *"unsupported metadata fails with a clear message instead of being ignored"* — before this, passing a bad key failed deep inside `cross_validate` with sklearn's generic error, after the whole search had already started.

## What changed

- Added a small up-front check in both `fit` methods. If the wrapped estimator's `fit` can't accept a metadata key, it raises a clear `TypeError` naming the parameter and the estimator, before the search runs.
- The check is conservative: if `fit` takes `**kwargs` (Pipeline, many meta-estimators) it can't know statically which keys are valid, so it skips and lets fit time decide. Routed keys like `clf__sample_weight` keep working.
- Tests: clear-message assertions for both estimators, plus a Pipeline case proving routed params aren't falsely rejected.
- Docs: a short "Passing Metadata" section in the CV guide covering `groups`/`sample_weight` and the limitations (Pipeline prefixes, and that full `enable_metadata_routing` isn't wrapped).
- CHANGELOG entry.

Full `test_genetic_search.py` + `test_feature_selection.py` pass (116). Backward compatible — valid usage is unchanged.

Note: this is the compatibility/clear-error slice, not full sklearn metadata routing (`enable_metadata_routing=True`), which would be a larger separate change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Full release notes with code examples are in the [documentation](https://sklearn
 
 ## Unreleased
 
+### New Features
+
+- `GASearchCV` and `GAFeatureSelectionCV` now reject unsupported `fit` metadata up front with a clear error naming the parameter and the wrapped estimator, instead of failing deep inside cross-validation. Estimators whose `fit` takes `**kwargs` (e.g. `Pipeline`) are left to validate at fit time, so routed keys like `clf__sample_weight` keep working (#364).
+
 ## 0.13.4
 
 ### New Features
--- a/docs-vitepress/versions/latest/guide/understand-cv.md
+++ b/docs-vitepress/versions/latest/guide/understand-cv.md
@@ -90,6 +90,34 @@ The `plot_fitness_evolution` chart shows:
 
 A healthy search shows the mean and best fitness rising together, then plateauing as the population converges.
 
+## Passing Metadata (groups & sample_weight)
+
+Both `GASearchCV` and `GAFeatureSelectionCV` forward metadata from `fit` through
+cross-validation and the final refit:
+
+```python
+from sklearn.model_selection import GroupKFold
+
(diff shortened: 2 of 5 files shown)


rodrigo-arenas:
Please add coverage for the missing _check_fit_params_supported branches so codecov/patch and codecov/project are green.

mayoka0:
I fixed the CHANGELOG.md conflict and kept both changes. I also added tests for the missing fit metadata branches. All 118 tests pass.

rodrigo-arenas:
Looks good from the implementation side; this is still blocked until an approving review is recorded and branch status is approved.

